PR #10, opened 2018-04-11
Add right side panel with favorite details and interpretations

tokland:
This PR implements the right side panel containing details of a favorite and its interpretations, similar to what we already have in d2-analysis. 

(cc: @adrianq, I also work with @EyeSeeTea)

I used [redux-first-router](https://github.com/faceyspacey/redux-first-router) for routing. Any action that should change the URL must go through the `loadFavorite` action creator.

Any comments/feedback to improve the code is welcome.


turban:
Hi, thanks for submitting this pull request - two comments:

1. This should be a separate component as the same functionality is needed in the new Data Visualizer app.
2. It would be niced to be noticed before you start on big pull requests like this - so we can sort out such issues before you start :-)

tokland:
Closing in favour of dhis2/d2-ui/pull/225

PR #17, opened 2018-09-18
DHIS2-4225 - Move layer card visibility toggle from header to footer.

amcgee:
From https://jira.dhis2.org/browse/DHIS2-4225

New position of visibility icon:

screenshot 2018-09-18 11 23 55

### DIFF
--- a/src/components/layers/basemaps/BasemapCard.js
+++ b/src/components/layers/basemaps/BasemapCard.js
@@ -118,7 +118,7 @@ const BasemapCard = props => {
                             opacity={opacity}
                             onChange={opacity => changeBasemapOpacity(opacity)}
                         />
-                        <div>
+                        <React.Fragment>
                             <Tooltip
                                 key="visibility"
                                 title={i18n.t('Toggle visibility')}
@@ -131,7 +131,7 @@ const BasemapCard = props => {
                                     {isVisible ? <VisibilityIcon /> : <VisibilityOffIcon />}
                                 </IconButton>
                             </Tooltip>
-                        </div>
+                        </React.Fragment>
                     </CardActions>
                 </CardContent>
             </Collapse>
--- a/src/components/layers/basemaps/BasemapCard.js
+++ b/src/components/layers/basemaps/BasemapCard.js
@@ -126,7 +126,6 @@ const BasemapCard = props => {
                                 <IconButton
                                     className={classes.visibility}
                                     onClick={toggleBasemapVisibility}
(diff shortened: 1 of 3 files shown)


src/components/layers/basemaps/BasemapCard.js
```diff
@@ -130,6 +118,20 @@ const BasemapCard = props => {
                             opacity={opacity}
                             onChange={opacity => changeBasemapOpacity(opacity)}
                         />
+                        <div>
```
amcgee: @turban I added this empty `<div>` wrapper because `<CardActions>` from material-ui assigns a `className` property to all children, but `<Tooltip>` throws doesn't support `className` and so throws an error when it is a first-level child of `<CardActions>`.  Let me know if you know of a better way to work around this.

amcgee: Actually, I just tested `React.Fragment` and it works, that will be better since it won't pollute the DOM.  I had previously tried the shorthand `<>` but it didn't work with our Webpack configuration.

amcgee: Updating the PR now

amcgee:
I used a [React Fragment](https://reactjs.org/docs/fragments.html) as a pseudo-element to avoid a console error when a Tooltip is assigned a className property a child of a CardActions, is this the right pattern?

src/components/layers/toolbar/LayerToolbar.js
```diff
@@ -75,6 +78,18 @@ const OverlayToolbar = ({
...
             onChange={opacity => onOpacityChange(layer.id, opacity)}
         />
+        <Tooltip
+            key="visibility"
+            title={i18n.t('Toggle visibility')}
+        >
+            <IconButton
+                className={classes.button}
+                onClick={() => toggleLayerVisibility(layer.id)}
+                style={{ backgroundColor: 'transparent' }}
```
turban: Remove style as we don't have it for the other buttons.

amcgee:
Updated to remove the style properties.  One quick question remaining for future reference:

> I used a [React Fragment](https://reactjs.org/docs/fragments.html) as a pseudo-element to avoid a console error when a Tooltip is assigned a className property as a child of a CardActions, is this the right pattern?

turban:
I think the pattern is fine - this seems more like a bug in Material UI. A cleaner fix would be to implement the toolbar in the same way as the overlay card.

PR #19, opened 2018-09-18
Update translation order (no-op)

amcgee:
The order of translation strings changed when we moved the visibility toggle, but the translation pot wasn't updated before PR #17 was merged.  This is because the translations are only generated in the prestart script.


amcgee:
(**NB** Rebased to remove the merge commit relics.)

PR #21, opened 2018-09-20
Implement layer card more menu and several code cleanup tasks.

amcgee:
This is groundwork for [DHIS2-3216](https://jira.dhis2.org/browse/DHIS2-3216).

I implemented a "more" menu in the Layer card toolbar to facilitate adding more options.  Before this change the toolbar was getting crowded, and the addition of `download` and `open in data visualizer` buttons was impossible.

I moved the `Data Table`, `Edit`, and `Remove` layer actions into the expansion menu, we need to decide whether `Edit` should remain a top-level toolbar action (Joe, @turban votes for keeping it top-level).

**Before**

screenshot 2018-09-20 12 12 51

**After**

screenshot 2018-09-20 12 25 05

In this change I've also performed several code cleanup tasks, including:
* Factor the Layer Toolbar to be used in both LayerCard and BasemapCard to reduce duplication
* Remove *.css files from the `src/components/layers` sub-tree, implementing those styles in JSS instead
* Fix highlighting of the currently selected basemap (id was not passed and so nothing was highlighted)

This PR does not yet implement the download functionality from DHIS2-3216 yet, so it should be merged separately.  Once it is merged, the PR for DHIS2-3216 will be concise and clean.

Another factorization task which I did not undertake would be to de-duplicate the code in `BasemapCard.js` and `LayerCard.js`, since they are very similar except for the Redux connections, but that is a task for another day.


turban:
Good work Austin and thanks for improving the code quality :-) We need the expansion menu for new fuctionality as data download  and switching view to charts/pivots. I'm not sure if we should move existing layer actions to this menu as they will be more hidden from the users, although it looks cleaner. Any opinions here Joe?

joeDHIS:
I think it makes sense to display the 'Edit' action in the layer toolbar **and** in the expansion menu. 

'Edit' is the most important command here so it's ok that we allow it to be accessed from two areas. We can see that the icon in the toolbar will be used by users who know their way around the app, whereas the 'Edit layer' in the expansion menu is great for helping users who are struggling to find out where to find that functionality.

The 'Data table' and 'Remove' actions should exist only in the expansion menu as these are secondary (and for remove, destructive) actions.

Toolbar actions can be displayed in this order:
Edit, Visibility Toggle (eye), opacity slider......overflow menu (aligned right)

turban:
Changes proposed are fine with me!

amcgee:
Thanks @joeDHIS, I'll make those changes!

amcgee:
PR updated, UI now looks like this (Note that the order of visibility toggle and opacity slider also changes for the Basemap card)

screenshot 2018-09-20 15 02 16

src/components/edit/EarthEngineDialog.js
```diff
@@ -70,10 +69,20 @@ const datasets = {
...
+        marginBottom: 12,
+    },
     legend: {
-        borderLeft: '12px solid #fff',
-        borderRight: '12px solid #fff',
+        padding: '0 16px 16px 32px',
+        margin: 0,
+        marginLeft: -32,
     },
+    // Unused?
```
amcgee: @turban I left this as as comment for now, feel free to have me remove if if you think it's safe to remove.

src/components/layers/basemaps/BasemapList.js
```diff
@@ -1,25 +1,35 @@
...
 import Basemap from './Basemap';
-import './BasemapList.css';
 
-const BasemapList = ({ id, basemaps, selectBasemap }) => (
-    <div className="BasemapList">
+const styles = {
+    container: {
+        maxHeight: 270,
+        overflowY: 'auto',
+        // overflowY: 'overlay', // This is a deprecated API and we can't have both 'overlay' and 'auto', see https://developer.mozilla.org/en-US/docs/Web/CSS/overflow
```
amcgee: Likewise with this comment.

turban:
The colors are not showing in the legend preview (top) - should be like bottom. 
skjermbilde 2018-09-21 kl 10 57 56
skjermbilde 2018-09-21 kl 10 57 46

src/components/edit/EarthEngineDialog.js
turban: The borders are adding extra margin to the "columns" - see screenshot here: https://github.com/dhis2/maps-app/pull/21#issuecomment-423465031

turban: Ref. https://stackoverflow.com/questions/20626685/better-way-to-set-distance-between-flexbox-items
Open for other ways to do the layout positioning :-)

turban:
Is it possible to add a tooltip to the opacity slider as well?

turban:
Missing text on layer thumbs without image. 

skjermbilde 2018-09-21 kl 11 10 04
skjermbilde 2018-09-21 kl 11 09 54

Better solution targeted for 2.31: https://jira.dhis2.org/browse/DHIS2-2566

amcgee:
PR Updated:

> The colors are not showing in the legend preview

Good catch!  This was because the Earth Engine Dialog was re-using the Legend class from the now-defunct Legend.css.  I refactored it to share style object


> The toolbar should be visisble when scroll is needed - not an issue with this PR - but we could fix it here :-)

I added a container div to the basemapCard so that scrolling works correctly, the issue was with `position: relative` on the basemap card.

> Missing text on layer thumbs without image.

This was an issue with the `lineHeight` property in JSS - it needs to explicitly specify `52px` not `52` since they have different meanings.  Fixed.

> Is it possible to add a tooltip to the opacity slider as well?

Yes, added `Set layer opacity` (with an additional wrapper div to make it appear consistently)

**One question**: I included a tooltip for the expansion menu button which just says `More`.  This isn't very helpful.  Should we remove it or change it to something more verbose and descriptive?  @joeDHIS  (screenshot below)

screenshot 2018-09-21 14 52 30

amcgee:
🎉 @joeDHIS for the More menu tooltip question above, we'll put that in a small separate change later on if we want to update it.

PR #22, opened 2018-09-20
Allow runtime selection of map ID and userOrgUnit ID in plugin.html

amcgee:
No changes to production code, just easier for testing.


turban:
Looks good! There are use cases of this when we need to add more config settings (full anlytical object), but then it's quick to comment out a few lines. We could also consider to keep this file out of the repo, it's just there for convenience.

PR #24, opened 2018-09-24
DHIS2-3216 - Enable layer data download as GeoJSON

amcgee:
Use [FileSaver.js](https://github.com/eligrey/FileSaver.js) package to save client-generated GeoJSON data.  Restrict download to `facility`, `boundary`, and `thematic` layers to prevent easy access to privacy-sensitive information (Events and Trackers).

Replace whitespace (`\s+`) in layer names with an underscore (`_`) before saving as `{layerName}.geojson`


turban:
We should support for events as well (but not when server clustering is used).

turban:
The PR looks fine - could we add some unit tests as well?

amcgee:
Yeah, will do!

amcgee:
> We should support for events as well (but not when server clustering is used).

I thought we wanted to disable that initially to obfuscate the ability to download raw event data?  Or was that only for `Tracked Entities` ?

turban:
Only for tracked enitities. Events are more anonymised.

jason-p-pickering:
@turban well, events may be more anonymized, but may still contain PII. It all depends on the design.  One could also argue that locations of facilities are also sensitive, so I am not really sure what the point of obfuscating anything is really?

turban:
@larshelge suggested this as a start. The malaria team would also like to download tracked entities. 

skjermbilde_2018-09-24_kl _13 55 48

jason-p-pickering:
I personally think its rather shallow to try and prevent people from downloading this. There are multiple ways to get around it, simply by making the API request, so not really sure why we are trying to prevent it. If its sensitive, then we need to enable a way to prevent people from accessing this information at all, since the response is inevitably going to be saved in the cache on the browser anyway right?

amcgee:
> I personally think its rather shallow to try and prevent people from downloading this. There are multiple ways to get around it, simply by making the API request, so not really sure why we are trying to prevent it. If its sensitive, then we need to enable a way to prevent people from accessing this information at all, since the response is inevitably going to be saved in the cache on the browser anyway right?

@jason-p-pickering I definitely agree with this in general - data access policies should be enforced server-side - but preventing zero-effort-required data access in the UI is a "security through obscurity" stopgap barrier we can put up against overt abuse in the meantime.  I also agree that Events might contain PII and so should be (1) more sensitive than Thematic aggregated data and (2) *usually* less sensitive than Tracked Entity data.  @larshelge should we be implementing low-friction privacy enhancements at the application level like this?  Worth a larger privacy/security policy discussion offline, I think.

**NOTE**

Since this PR has other things in it too - we're talking about **allowing full data download as GeoJSON for different layer types**.  As it stands I'm allowing it for all Thematic, Boundary, and Facility layers (the same ones for which you can currently open a DataTable to see the data visually) as well as for Event layers which do not have Server Side Clustering.  I am not allowing data download for Tracked Entities layers.  

The most immediate question is - **should the maps app disable data download for Event layers?**

turban:
Maybe we should use "Download GeoJSON" rather than "Download data" since we don't (yet) support other formats like CSV?

package.json
```diff
@@ -49,6 +49,7 @@
         "d3-color": "^1.2.3",
         "d3-format": "^1.3.2",
         "d3-time-format": "^2.1.3",
+        "enzyme-to-json": "^3.3.4",
```
jenniferarnesen: Is this used anywhere?

src/components/layers/toolbar/__tests__/LayerToolbar.spec.js
```diff
@@ -0,0 +1,138 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { LayerToolbar } from '../LayerToolbar';
+
+describe('LayerToolbar', () => {
+    it('Should render only a visibility toggle and opacity slider', () => {
+        const wrapper = shallow(
```
jenniferarnesen: You could save some lines of code by wrapping the shallow in a function. Here is an example:
https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/components/Visualization/__tests__/BlankCanvas.spec.js

package.json
amcgee: Now it is!  I missed the update in the jest config before, good catch.  Just pushed the fix.

`enzyme-to-json` translates the Enzyme shallow DOM rendering structure into JSON for use in Jest snapshot tests, so it looks much cleaner and can be easily diffed in (for instance) `LayerToolbar.spec.js.snap`

PR #25, opened 2018-09-24
Browser issues fixes

turban:
This PR fixes various browser issues: 
- .json removed from analytics requests (in maps app, not yet in file menu dependency)
- Remove object.assign, for of loops and string.includes for IE11 support
- IE11 style fixes
- Totals/details radio toggle fix after MUI upgrade (not browser related)


amcgee:
I don't think we should change all our uses of `Array.prototype.includes`, but rather we should upgrade Webpack to a version that supports it (version > `4.2.0`, the latest being `4.19.1`, see [this](https://github.com/webpack/webpack/issues/6768) and [this](https://github.com/webpack/webpack/pull/6739) and [this](https://github.com/webpack/webpack/issues/6764).  We should figure out if we'll have any issues upgrading to 4.x, I doubt we will but it's possible.

Similarly, I think we can support `Object.assign` easily enough just by including `babel-polyfill`, which is already in our `package.json`.

I think this would simplify the change a lot and make it easier to ensure we don't break IE going forward.  Thoughts?

amcgee:
Also, why don't you leave `i18n/en.pot` out of this change since it's not directly related and will conflict with the changes in #24 (which also includes `Set layer opacity`)

turban:
I agree, let's upgarde to webpack 4 and babel 7 (separate issue) - and then we can continue on this issue afterwards.

amcgee:
Note that `babel-polyfill` will probably add to our download size, we could pick out just the `Object.assign` bit if we want it, or we could decide not to use `Object.assign`

amcgee:
And one more - it looks like we're already including `babel-polyfill` as a vendor script in `webpack.config.js` so I'm not sure why we're not able to use `Object.assign` ...

turban:
This PR was replaced by https://github.com/dhis2/maps-app/pull/26

PR #31, opened 2018-10-09
d2-ui-file-menu upgrade

turban:
Partly fixes: https://jira.dhis2.org/browse/DHIS2-4482

Just a dependency upgrade of d2-ui-file-menu. The changes in devDependencies only seems to be reordering.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "maps-app",
-    "version": "31.0.6",
+    "version": "31.0.7",
     "description": "DHIS2 Maps",
     "main": "src/app.js",
     "repository": {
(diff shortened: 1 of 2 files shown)


amcgee:
@turban Did you see my review comment about using `^` for dependency versioning?  I think the file-menu upgrade should maybe have been a minor version bump though, so perhaps we have to use 3.2.1 explicitly...

turban:
Missed that one. Now changed to "@dhis2/d2-ui-file-menu": "^3.2.1" in package.json in master. Will keep an eye on what it resolves to - I agree it should have been a minor version bump.

PR #33, opened 2018-10-11
Support multiple org unit roots

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-4910

Org unit trees can sometimes have multiple org unit roots, and this PR will create one tree per root using the d2-ui OrgUnitTreeMultipleRoots component.


amcgee:
Looks good!  I'm curious why there are two separate components but that's a question for another day...

PR #34, opened 2018-10-15
Geometry type fix for TEI layer

turban:
This PR fixes a problem with geometry types in TEI layer. Polygons have previously been stored wrong in DHIS2 backend, but this is no longer the case.


turban:
>  I presume the backend change has been back-ported to 2.30 as well?

Seems like, but will ask @stian-sandvold before backporting.

PR #36, opened 2018-10-18
DHIS2-4857 - Support event layer downloads with multiple Feature Key formats.

amcgee:
This turned into a pretty big pull request - I will submit it for review as is and make modifications to the MetaData output format options in another change once this has been merged.  The major changes here are:

* Add `DataDownloadDialog` for downloading layer data
* Support customization of event layer GeoJSON key formats
* Show a loading spinner while downloading Event layer data at download-time (data is not available locally so it must be requested from the server)
* Show an error message in the dialog if the download fails for some reason
* Use a `redux-observable` epic to manage the download asynchronous behavior

**Tangential**

* Refactor `eventLoader.js` for less repetition and more reusability
* Move helper functions to `util/geojson.js` to keep `eventLoader`, `epics/dataDownload`, and `DataDownloadDialog`
* Implement unit tests with 100% coverage for `util/geojson.js`
* Add `Netlify.toml` to automatically deploy branches and PRs, like [this](https://deploy-preview-36--dhis2-maps.netlify.com/) for this PR.  Once merged this will be available on Master and on all PRs to Master
* Remove custom styling from Maps dialogs and implement DHIS2 identity-compliant Dialog styles in the dhis2 MUI theme so that they are automatically adopted in any MUI dialog.  **TODO**: move this to `d2-ui` and remove customization from the various dialogs in `d2-ui` components.

Dialog now looks like this for Event and Thematic layers, respectively (showing error state for reference):

screenshot 2018-10-29 14 56 19
screenshot 2018-10-29 14 57 03

**NB** Comment updated and WIP title removed, requesting review from @turban and @joeDHIS - you can test these changes [here](https://deploy-preview-36--dhis2-maps.netlify.com/) or check them out locally.


amcgee:
Hey @joeDHIS and @turban, I'd love some feedback on this Event Download implementation!

- I've implemented a dialog for all data downloads, including a link to [geojson.org](http://geojson.org) for some information on geojson (@turban should we link somewhere else?).
- For Event layers I've allowed the user to select the output format (ID, Code, or Name) for data dimensions and to select Human Readable (Name) or ID for all other GeoJSON keys.  This works fine, but it's a little clunky and inconsistent across different layer types.  I'm not sure how to get the ID/Code/Name from the webapi for other layer types (if it's even possible) and it would require a bit more refactoring to make that work in the code.

So three things I could use some help with:
1) Feedback on the dialog content, loading behavior, and output formats
2) For longer blocks of text like this, how do we usually do translation?  I think we probably don't want to inline the full text as the translation key?
3) Since the OutputIdScheme parameter appears to be broken in the backend I had to work around it here, feedback on the workaround would be great too. @janhenrikoverland @edoardo 

I'm also testing continuous deployment of branches and PRs, so **you can see a working version of this change** (pointing at the play/dev server) [here](https://deploy-preview-36--dhis2-maps.netlify.com/).  It will be automatically updated when the PR is changed, and you can also get to the link from the "netlify/dhis2-maps/deploy-preview" Check below.  NB: I had to manually add this url to the CORS whitelist but otherwise everything was automatic, yay!  👍 

Thanks!

amcgee:
Here's a screen recording for those who don't want to visit the [live demo](https://deploy-preview-36--dhis2-maps.netlify.com/)

maps-36-demo

amcgee:
And another to show the difference when downloading a thematic layer (in which case we already have the data locally, so no download delay, but also don't fetch updated key names so there's no option for the user there.

maps-36-demo2

turban:
Nice work! 

geojson.org is not very descriptive of the standard. I would suggest to remove the link and write something like: 

"The data for this layer will be downloaded in GeoJSON format. This format is supported my most mapping programs like QGIS and ArcGIS."

We could later link to a tutorial in our docs about how to import and use data in QGIS.

turban:
Small bug: "More actions" tooltip is displayed above the download dialog.  

skjermbilde 2018-10-24 kl 14 59 16

src/loaders/eventLoader.js
```diff
@@ -188,23 +163,73 @@ export const getAnalyticsRequest = async (
...
+    headers,
+    names,
+    event,
+    eventCoordinateField
+) => {
     const properties = event.reduce(
         (props, value, i) => ({
             ...props,
-            [headers[i].name]: names[value] || value,
+            [names[headers[i].name] || headers[i].name]: value,
```
amcgee: @turban I believe the code was incorrect here before since names comes from the header list. I updated it to work for Event Download but would love your review to make sure I didn't change the behavior of this function.

turban:
Bug: Add event layer -> Program: Inpatient morbidity and mortality -> Add layer -> 

TypeError: Cannot read property 'outputIdScheme' of undefined

joeDHIS:
There are some guidelines that have been created for d2ui, which can be seen [here](https://invis.io/A7LT4TJYETS#/324707554_molecules). They haven't been built yet though. Looking at the examples, the "Add new Event Layer' looks correct.

Everything else looks nice. One question - do we want to inform the user why there is no option for setting key names for thematic layers? It's not immediately obvious to me, but I don't have any experience with GIS data, so maybe it's something our expected audience will understand. If not, a simple "Format for GeoJSON features keys cannot be modified for thematic layers", could suffice. That way a user isn't searching around, thinking they can't find the option or something. I'm happy to defer to you guys on this one and trust what you think.

turban:
Comparsion of the different GeoJSON-formats provided for event layer download:

1. Meta-data ID Format: ID
id

2. Meta-data ID Format: Code
code

3. Meta-data ID Format: Name
name

4. Meta-data ID Format: Name + human-readable keys
name_readable

I think 4 is the most readable and user friendly format - and I would suggest it to be the default.

@joeDHIS are we using Title Case or Sentence case for text field labels, titles etc.?

turban:
The text field is very long and the content is narrow - better to set a shorter max length than 100% or make the dialog smaller? 

skjermbilde 2018-10-24 kl 16 13 53

@joeDHIS are we using Title Case or Sentence case for text field labels, titles etc.?

src/loaders/eventLoader.js
```diff
@@ -65,30 +49,16 @@ const eventLoader = async layerConfig => {
...
-            .map(row =>
-                createEventFeature(
-                    response.headers,
-                    names,
-                    row,
-                    eventCoordinateField
-                )
-            )
-            .filter(feature => isValidCoordinate(feature.geometry.coordinates));
+        const { names, data, response } = await loadData(analyticsRequest);
```
turban:
```suggestion
        const { names, data, response } = await loadData(analyticsRequest, config);
```

amcgee:
Good catch , yeah that is the change @turban thanks!

amcgee:
I matched the text casing in the Pivot Table app download menu

amcgee:
Feedback welcome and appreciated on styling, I haven't taken the time to fine-tuned it yet so there are inconsistencies.

amcgee:
> Small bug: "More actions" tooltip is displayed above the download dialog.
> 
> skjermbilde 2018-10-24 kl 14 59 16

Yeah, it's an annoying bug but unrelated to this change.  Something to do with the combination of tooltip and popup menu, when the menu is dismissed the tooltip reappears until you enter/exit the hover state again.

amcgee:
Unit test coverage of `util/geojson.js` file:

screenshot 2018-10-30 17 08 52

turban:
Should we add tests for new React components like DataDownloadDialog as well?

turban:
skjermbilde 2018-10-31 kl 16 28 30

Some suggestions: 
- A bit more "air" above the select box
- "Please select the format for GeoJSON properties"
- "ID-format" (skip Meta-data...)
- "Use human-readable keys" (less accurate, but I guess most users don't know the meaning og non-dimension attributes).

amcgee:
Thanks for the input!

> * Increase spacing above the select field

Done

> * "Please select the format for GeoJSON properties"

I updated it to "Please select the format for dimension attributes", think it's a little more clear

> * "ID-format" (skip Meta-data...)

I changed this to `Meta-data ID sheme` which matches the download option in the Pivot Table and new Data Visualizer download menus.  If we want to update it (and I agree it is not an accessible name), we should do it everywhere.

screenshot 2018-10-31 17 29 56

> * "Use human-readable keys" (less accurate, but I guess most users don't know the meaning of "non-dimension attributes").

I changes this to `Use human-readable keys for all other attributes` since I added `dimension` above.  Agree it's not the most user-friendly but it's important to differentiate the two in some way, otherwise both inputs seem like they do the same exact thing.  This text may change or go away when we have server-side support for outputIdScheme.

amcgee:
Great suggestions!  Updated with those minor changes.

joeDHIS:
I'm taking a look at the dialog now - functionality looks good, I may make a few visual changes to make the info clearer. Will post any changes/specs here!

amcgee:
Thanks @joeDHIS, two follow-ups:

1) For the title "GeoJSON keys" do you think this is descriptive enough for the user?  I find it ambiguous to have "GeoJSON keys ID format" and "Use human-readable keys" (they sound like the same thing) which is why I called them "dimension attributes" and "all other attributes" explicitly.  Again this may go away soon but worth pointing out.

2) As stated above I like `ID Format` but it's inconsistent with other apps which have a similar feature, where this is called obtusely `Meta-data ID scheme`, should we be clear or consistent here?  I think if we update one we should update all of them.

amcgee:
If we want to update the name I will do the same in PRs to the other apps.

amcgee:
Made the changes requested by @joeDHIS, dialog screenshots below.  Let's get this merged then address the above questions.

screenshot 2018-11-01 13 02 50
screenshot 2018-11-01 13 03 06

amcgee:
💯

joeDHIS:
> For the title "GeoJSON keys" do you think this is descriptive enough for the user? I find it ambiguous to have "GeoJSON keys ID format" and "Use human-readable keys" (they sound like the same thing) which is why I called them "dimension attributes" and "all other attributes" explicitly. Again this may go away soon but worth pointing out.
I think we need to get some user feedback on this, potentially Scott is the one who could help us. I'm unsure how technically adept the average Maps user is (of course, there is no real average, but we have to land somewhere). 

> As stated above I like ID Format but it's inconsistent with other apps which have a similar feature, where this is called obtusely Meta-data ID scheme, should we be clear or consistent here? I think if we update one we should update all of them.
I think it is best to be clear first and consistent second, otherwise we risk just making the problem even worse by extending the use of weird terms. Let's go forward with ID format and make a point of updating it elsewhere too. (This will be easier said than done, but a good goal.)

Thanks for updating from the specs so fast! 👍

PR #37, opened 2018-10-31
PNG export and map legend refactoring

turban:
This PR allows the user to download maps as PNGs from DHIS2 Maps. This is done in the browser which has some limitations:

1. It is not working all browsers (see below)
2. The maps are not accessible through the Web API

The solution is based on similar functionality from [kepler.gl](http://kepler.gl/) and [dom-to-image](https://github.com/tsayen/dom-to-image).

Browser support: [Chrome and Firefox (not IE and Safari)](https://github.com/tsayen/dom-to-image#browsers). The PR includes a check if the browser supports download or not. 

A "Download" button is added to the top menu, and a dialog opens on click:

skjermbilde 2018-10-31 kl 17 45 21

The map goes into "download mode" when the dialog is open. This hides the map controls and adds a legend. The user can change the legend position. 

The PNG downloaded will have the same size and contents as shown in the app. 

The file name is currently "map-{random}.png". Open for other suggestions. Remember that a map can be downloaded before it is saved (and have a name and id).  

The PR also includes a refactor of map legend rendering. We now support legends in 3 positions: Layers panel, on the map (download), and in a "control" (dashboard). The same react components are used for all 3. 

JIRA-issue: https://jira.dhis2.org/browse/DHIS2-245

Unit tests are still missing. Would be nice to get some initial feedback before I start adding tests.


src/actions/download.js
```diff
@@ -0,0 +1,25 @@
+import * as types from '../constants/actionTypes';
+
+// Set download state (download dialog is open)
+export const setDownloadState = isActive => ({
```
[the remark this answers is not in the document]
turban: Fixed

src/actions/download.js
```diff
@@ -0,0 +1,25 @@
...
+    type: types.DOWNLOAD_STATE_SET,
+    payload: isActive,
+});
+
+// Whether the name should be included on the map download
+export const setDownloadNameState = show => ({
```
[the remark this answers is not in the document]
turban: Fixed

src/actions/download.js
```diff
@@ -0,0 +1,25 @@
...
+    type: types.DOWNLOAD_NAME_STATE_SET,
+    payload: show,
+});
+
+// Whether the legend should be included on the map download
+export const setDownloadLegendState = show => ({
```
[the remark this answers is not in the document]
turban: Fixed

src/components/layers/legend/Legend.js
```diff
@@ -87,4 +87,4 @@ Legend.propTypes = {
     sourceUrl: PropTypes.string,
 };
 
-export default withStyles(styles)(Legend);
+export default withStyles(styles)(MapLegend);
```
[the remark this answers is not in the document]
turban: I've renamed the component to "Legend" which is the generic one, and used by the  MapLegend component. Also adding an explanation to differentiate them.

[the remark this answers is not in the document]
turban: MapLegend component has been renamed to DownloadLegend and moved to the download folder.

src/components/map/MapLegend.js
```diff
@@ -0,0 +1,75 @@
...
+        backgroundColor: 'rgba(255,255,255,0.9)',
+        boxShadow: theme.shadows[1],
+        borderRadius: theme.shape.borderRadius,
+        fontSize: theme.typography.fontSize,
+    },
+    title: {
+        fontSize: 15,
+        fontWeight: 'normal',
+        margin: 0,
+        paddingBottom: 10,
```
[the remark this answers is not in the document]
turban: Fixed

src/components/map/MapName.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+
+const styles = theme => ({
+    name: {
+        position: 'absolute',
+        top: 8,
```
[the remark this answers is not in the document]
turban: Fixed

src/components/map/MapLegend.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+import Legend from '../layers/legend/Legend';
+
+const positions = showName => ({
```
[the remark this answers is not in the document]
turban: Fixed

src/components/map/MapLegend.js
```diff
@@ -0,0 +1,75 @@
...
+    bottomright: {
+        bottom: 20,
+        right: 8,
+    },
+});
+
+const styles = theme => ({
+    root: {
+        position: 'absolute',
+        padding: '8px 8px 0 8px',
```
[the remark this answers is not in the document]
turban: Fixed

turban:
Thanks for your feedback @amcgee - your proposed changes are now implemented in this PR.

I see your concern about having both "Download" in the main menu and "Download data" for individual layers. I think the main concern is that users who wants to download data will look at the wrong place. Maybe we can add a short text in the Download dialog:

"Here you can download a PNG image of your map. You can download data in the "More actions" menu in the layers panel."

Any suggestions here @joeDHIS?

joeDHIS:
> I see your concern about having both "Download" in the main menu and "Download data" for individual layers. I think the main concern is that users who wants to download data will look at the wrong place. Maybe we can add a short text in the Download dialog:
> 
> "Here you can download a PNG image of your map. You can download data in the "More actions" menu in the layers panel."
> 
> Any suggestions here @joeDHIS?

My first thought was to rename to "Save" - as in "Save map as a .jpg", but I think that gets messy with "Saving/Starring/Favoriting". 

We could stick with 'Download' and potentially solve this by allowing a second route into the 'download individual layer' dialog. See attached mockup where the 'Download' option from the menu actually does give 'Download' access to everything. I think then it makes more sense that the menu item is called download, since that's what it is.

download_combo_menu

src/components/download/DownloadDialog.js
```diff
@@ -0,0 +1,169 @@
...
+                        </Fragment>
+                    ) : (
+                        i18n.t('Map download is not supported by your browser. Try Google Chrome or Firefox.')
+                    )}
+                </DialogContent>
+                <DialogActions>
+                    <Button color="primary" onClick={this.onClose}>
+                        {isSupported ? i18n.t('Cancel') : i18n.t('Close')}
+                    </Button>
+                    {isSupported &&
```
amcgee: As written I think it could be possible for an error to occur, displaying 'Map download is not supported by your browser. Try Google Chrome or Firefox' but still showing an enabled 'Download' button.  It probably wouldn't enter that state in the wild very often, but we should probably use the same conditional (`isSupported && !this.state.error`) everywhere (line 116 and 118 here)

src/components/download/DownloadLegend.js
```diff
@@ -0,0 +1,88 @@
...
+        </div>
+    );
+};
+
+DownloadLegend.propTypes = {
+    position: PropTypes.oneOf([
+        'topleft',
+        'topright',
+        'bottomleft',
+        'bottomright',
```
amcgee: Minor: can this share the `positions` const array from LegendPosition so we're not defining that list multiple times?

src/components/download/__tests__/DownloadButton.spec.js
```diff
@@ -0,0 +1,47 @@
...
+                .find('WithStyles(Button)')
+                .render()
+                .text()
+        ).toEqual('Download');
+    });
+
+    it('Should render a download dialog', () => {
+        const wrapper = renderComponent();
+        expect(
+            wrapper.find('Connect(WithStyles(DownloadDialog))').exists()
```
amcgee: This could be extended to check if passing different props shows and hides the DIalog.  With Enzyme you can inspect props with `wrapper.find('something').prop('open')` and test that it is `false` by default and `open` when a different prop is passed or the button is clicked (depending on whether or not Redux is involved)

src/components/download/__tests__/DownloadLegend.spec.js
```diff
@@ -0,0 +1,83 @@
...
+            />
+        );
+
+    it('Should render a legend component', () => {
+        expect(renderComponent().exists()).toBe(true);
+    });
+
+    it('Should set legend position class', () => {
+        const wrapper = renderComponent();
+        expect(wrapper.hasClass('bottomright-class')).toBe(true);
```
amcgee: Minor: these should probably reference `classes.bottomright` etc. instead of hard-coding the class name.

amcgee: Same with this whole file

src/components/map/Map.js
```diff
@@ -27,11 +30,21 @@ const layerType = {
     external: ExternalLayer,
 };
 
+const styles = {
```
amcgee: Since we're introducing `styles` here can we remove the inline styles from this file at the same time instead of keeping both?

src/components/map/Map.js
```diff
@@ -152,20 +170,40 @@ class Map extends Component {
...
-                            key={config.id}
-                            index={index}
-                            openContextMenu={openContextMenu}
-                            {...config}
-                        />
-                    );
-                })}
-                <Layer key="basemap" {...basemapConfig} />
+            <div
+                className={isDownload ? 'dhis2-maps-download ' : null}
```
amcgee: (from above) Why is `dhis2-maps-download` a global css style?

src/components/map/plugin/PluginMap.js
```diff
@@ -199,44 +186,51 @@ class PluginMap extends Component {
...
-                    onDrillUp={() => this.onDrill('up')}
-                />
-            </div>
-        ) : (
-            <div style={styles.alerts}>
-                {alerts.map(alert => (
-                    <div key={alert.id} style={styles.alert}>
-                        <strong>{alert.title}</strong>: {alert.description}
+        return (
+            <MuiThemeProvider theme={muiTheme}>
```
amcgee: Hmmm... I'm not actually sure we want to include the theme in `PluginMap`.  There will be multiple instances if the parent app (dashboards) has one, and the parent won't be able to override anything in the theme that affects the map.  But maybe we want to explicitly theme it here?

src/epics/orgUnits.js
```diff
@@ -37,7 +37,7 @@ export const loadOrgUnitTree = action$ =>
         getD2()
             .then(d2 =>
                 d2.models.organisationUnits.list({
-                    level: 1,
+                    userDataViewFallback: true,
```
amcgee: What does this change do?

src/components/download/DownloadDialog.js
turban: Good catch! Added: 
const isSupported = downloadSupport() && !this.state.error;

src/components/download/__tests__/DownloadButton.spec.js
turban: This is checked in DownloadDialog.spec.js

src/components/download/__tests__/DownloadLegend.spec.js
turban: Fixed

src/components/map/Map.js
turban: Global css style replaced with styles object.

src/components/map/plugin/PluginMap.js
turban: Removed theme again.

src/epics/orgUnits.js
turban: Fixes a bug reported by @janhenrikoverland. There might be users that don't have access to level 1 in the org unit hierarchy, and this query should only return the topmost the user has access to. Added a comment in the code explaining it. This fix was also backported to 2.30 and 2.29.

PR #39, opened 2018-11-07
DHIS2-4917 - Support start/end date periods for thematic layers

amcgee:
There's a lot of duplication and inline style usage in EventDialog, ThematicDialog, and TrackedEntityDialog - we should refactor them.  For now I added support for start/end periods to Thematic layers and added error handling to Event and TrackedEntity layers which previously would break silently when an invalid date was entered.


turban:
We should also check if end date is later that start date (web api error if not)

amcgee:
Pushed another minor update to remove the default value (malformed string, was never used and caused a console warning).  Ready to merge if that minor change is approved

amcgee:
There's also a theme issue causing large body texts which I will fix separately

amcgee:
Going to go ahead and merge since it was a minor change 👍

PR #41, opened 2018-11-08
Upgrade d2-ui dependencies and use the theme file from d2-ui-core

amcgee:
This should be a one-for-one replacement and I haven't noticed any regressions in the FileMenu, Interpretation, or OrgUnit components


amcgee:
We're going to push a new version of `d2-ui` with a very slightly updated theme, so I will wait until that is published before updating this PR and merging

amcgee:
Going ahead with this, will upgrade when we finally get a new published d2-ui

PR #44, opened 2018-11-09
DHIS2-5171 - Add period label to EarthEngine layer cards.

amcgee:
Some examples:

screen shot 2018-11-09 at 1 36 40 pm

---

screen shot 2018-11-09 at 1 41 13 pm

---

(elevation has no period options)

screen shot 2018-11-09 at 1 42 19 pm


turban:
I like the screenshot and the purpose of this PR - have you tested that it works when you open a saved map with an EarthEngine layer?

amcgee:
@turban as it stands it will not, that is an excellent point.

turban:
We could merge this PR now, and then look at a way to store the period name in the mapView in a separate PR.

turban:
Yes, there is a third for next week :-)

turban:
New commit to this PR: 
- rename from periodLabel to periodName (as it might be used for different stuff)
- Storing periodName in mapView config for EE layers

PR #47, opened 2018-11-14
Same org unit tree style as data viz app

turban:
Added the same styles for the org unit tree as im the Data visualizer app, to keep the design consistent: 

skjermbilde 2018-11-14 kl 12 15 52

These styles should be shared among the apps, and it should be implemented using classes rather than inline styles. Out of scope of this PR.


turban:
Included a style fix for map download in this PR: Remove blue link style in the map legend

src/components/orgunits/OrgUnitTree.style.js
```diff
@@ -0,0 +1,44 @@
+// TODO: The d2-ui component should use classes, not styles!
+// Copied from https://github.com/dhis2/d2-ui/blob/master/packages/org-unit-dialog/src/styles/OrgUnitDialog.style.js
```
amcgee: I don't love the idea of copying this file, it will get out of sync.  Is it an exact copy?  Could we instead `import orgUnitStyles from '@dhis2/d2-ui-org-unit-dialog/styles/OrgUnitDialog.style'` ?  I know that adds a dependency, though, so it might not be a great solution.  I imagine it would take a lot of work to actually use the `d2-ui` component?

turban: Yes, it will be a lot of work to adapt the d2-ui OrgUnitDialog component as we need different setups for different layer types. 

I agree that copying the styles is not ideal, but doing it for now. When we move away from Material UI for these componets, we need to get a better solution. 

As the Maps app bundle is already quite huge, I would prefer not adding an extra dependency just to include the styles.

amcgee: I believe that using the import syntax I suggested it should only pull that single file into the bundle, but that should be tested.  If that's not high-priority we can duplicate for now.

amcgee:
Confirmed that the new import increases the bundle size by only 2kb (pulling in only that single file)  Approved and merging!

screen shot 2018-11-15 at 2 43 10 pm

(left is after change)

screen shot 2018-11-15 at 2 39 10 pm

PR #49, opened 2018-11-14
Thematic as first layer item in add layer dialog

turban:
Small fix that sets thematic layer (most used) first in the add layer dialog:  

skjermbilde 2018-11-14 kl 15 19 29

https://jira.dhis2.org/browse/DHIS2-4224

### DIFF
--- a/src/reducers/layers.js
+++ b/src/reducers/layers.js
@@ -5,7 +5,7 @@ const defaultLayers = [
         layer: 'thematic',
         type: 'Thematic',
         img: 'images/thematic.png',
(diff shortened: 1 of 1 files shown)


turban:
Included a fix for https://jira.dhis2.org/browse/DHIS2-4226 in this PR. The default layer opacity for thematic layers is increased to 0.9. Still keeping them a bit transparent as the layers are added on top of each other, which might hide underlying layers totally if no transparency. 

skjermbilde 2018-11-14 kl 15 26 29

PR #50, opened 2018-11-14
Keep map view on layer update

turban:
This PR will only zoom/pan the map the first time a layer is added, and not on every update.

### DIFF
--- a/src/components/map/ThematicLayer.js
+++ b/src/components/map/ThematicLayer.js
@@ -10,6 +10,7 @@ class ThematicLayer extends Layer {
             labels,
             labelFontSize,
             labelFontStyle,
+            editCounter,
         } = this.props;
 
         const map = this.context.map;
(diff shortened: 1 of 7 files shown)


src/components/map/ThematicLayer.js
```diff
@@ -34,11 +35,9 @@ class ThematicLayer extends Layer {
         this.layer.on('click', this.onFeatureClick, this);
         this.layer.on('contextmenu', this.onFeatureRightClick, this);
 
-        const layerBounds = this.layer.getBounds();
-
-        if (layerBounds.isValid()) {
-            map.invalidateSize();
```
amcgee: Is `map.invalidateSize()` no longer necessary?

turban: Yes, I did some tests, and it seems that it's not needed - and it's not used for the other layer types. map.invalidateSize() should be used when the size of the map div is changed, which is not the case here.

amcgee:
One edge case for this issue - what should happen if a layer is added, then removed, then another layer added?  Should the map reset its bound in that case?

turban:
Hmm - if a layer is removed - and another is added, the map will zoom to the contents of the new layer.  I think this is consistent. There is also a zoom to content button on the map, which will adjust to the content across all layers.

PR #52, opened 2018-11-15
Remove SASS styles

turban:
This PR removes SASS styles and dependencies from the Maps app. Most of the SASS styles were no longer in use.


webpack.config.js
```diff
@@ -68,7 +68,7 @@ const webpackConfig = {
             },
             {
                 test: /\.scss$/,
```
amcgee: Do we need an `scss` loader at all now?

turban: good catch! scss loaded removed.

PR #54, opened 2018-11-19
Style and bug fixes

turban:
This PR fixes various issues:
- Upgrade GIS API with the lastes Google Maps API version
- Convert font with 'px' size to integer
- Use className rather than inline styles 
- Close download dialog on image download
- Output download error to console (temporarily)
- Avoid horisontal scroll in layer dialogs
- Context menu style fixes (align icon and labels) 
- Fix map control styles (due to new global stylesheet from ui/core)

### DIFF
--- a/src/components/orgunits/OrgUnitDialog.js
+++ b/src/components/orgunits/OrgUnitDialog.js
@@ -26,6 +26,7 @@ const styles = {
     data: {
         float: 'left',
         width: 345,
+        marginTop: -12,
     },
     nodata: {
         fontStyle: 'italic',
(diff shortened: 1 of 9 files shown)


src/components/orgunits/OrgUnitDialog.js
```diff
@@ -141,7 +143,7 @@ export class OrgUnitDialog extends Component {
                             periodType={periodType}
                             period={period}
                             onChange={this.onPeriodChange}
-                            style={{ height: 70, margin: '-17px 0 0 3px' }}
+                            // style={{ height: 70, margin: '-12px 0 0 3px' }}
```
amcgee: Can we remove this instead of commenting or is it a placeholder for later?

turban: Ah, yes it should be removed. Fixing and merging.

PR #56, opened 2018-11-21
Task/ie11 fixes

amcgee:
This makes the application functional in IE11 by removing Object.assign and Array/String.includes calls as well as polyfilling the URL object (Which caused `Object does not support this action` errors in FileMenu.

Much credit to @turban for the first version of the code changes in #25 

I noticed several styling inconsistencies as well but will open those in a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
         "redux-thunk": "^2.3.0",
         "rxjs": "5.5.6",
         "ui": "github:d2-ci/ui",
+        "url-polyfill": "^1.1.0",
         "whatwg-fetch": "^3.0.0"
     },
     "devDependencies": {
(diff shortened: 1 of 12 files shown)


amcgee:
Question: Should the polyfill be included in `map.js` also?

amcgee:
Question: Should the polyfill be included in `map.js` also?

amcgee:
> > Should the polyfill be included in map.js also?
> 
> I guess it should if needed :-)
> We could test first using plugin.html - I'm not sure if we use Object.assign and Array/String.includes in map.js and includes (we should check in loaders/ and components/map/

It's actually the use of `new URL` somewhere in FileMenu that requires this polyfill, we should be fine for Object.assign and Array/String.includes

PR #57, opened 2018-11-21
Bug fixes for empty dimensions and period validation

turban:
This PR fixes two bugs: 
- Invalid data dimensions (dx) with empty item array is not passed on to web api
- Older map objects (filters) includes dates with hours, and this fix only checks the validity of the date part which we currently support in the event dialog.

### DIFF
--- a/src/loaders/eventLoader.js
+++ b/src/loaders/eventLoader.js
@@ -219,6 +219,6 @@ const getFilterOptionNames = async (filters, headers) => {
 
 // Empty filter sometimes returned for saved maps
 const isValidDimension = ({ dimension, items }) =>
-    Boolean(dimension && !(items && !items.length));
+    Boolean(dimension && (!items || items.length));
 
 export default eventLoader;
--- a/src/loaders/eventLoader.js
+++ b/src/loaders/eventLoader.js
@@ -142,7 +142,7 @@ export const getAnalyticsRequest = async ({
 
     if (dataItems) {
         dataItems.forEach(item => {
-            if (item.dimension) {
+            if (isValidDimension(item)) {
                 analyticsRequest = analyticsRequest.addDimension(
                     item.dimension,
                     item.filter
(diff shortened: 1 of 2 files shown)


src/loaders/eventLoader.js
```diff
@@ -217,4 +217,8 @@ const getFilterOptionNames = async (filters, headers) => {
     );
 };
 
+// Empty filter sometimes returned for saved maps
+const isValidDimension = ({ dimension, items }) =>
+    Boolean(dimension && !(items && !items.length));
```
amcgee: Is `{dimension: 'something', items: null}` valid?  If so, maybe this would be a little cleaner without the double negation as `dimension && (!item || items.length)` (though maybe not any better, they're both a little hard to read) and if not then it should be `dimension && items && items.length`

src/util/helpers.js
```diff
@@ -125,8 +125,8 @@ export const formatTime = date =>
     timeFormat(DATE_FORMAT_SPECIFIER)(new Date(date));
 
 export const getStartEndDateError = (startDate, endDate) => {
-    const start = parseTime(startDate);
-    const end = parseTime(endDate);
+    const start = parseTime(startDate.substring(0, 10)); // Only check date part
```
amcgee: What is the issue we're trying to solve here?  This check was introduced to catch invalid input like `01/12/yyyy`, so the time part shouldn't matter unless I'm missing something.  When is time input allowed?

src/loaders/eventLoader.js
turban: Changed to more readable `dimension && (!item || items.length)` :-)

src/util/helpers.js
turban: Keeping this after our discussions. The issue is that the backends adds the hour/minutes/seconds to periods - and I think we should keep this to be more future proof.

PR #58, opened 2018-11-21
Allow empty style by data element

turban:
This fix allows to empty the "Style by data element" select box.


src/components/dataItem/DataItemSelect.js
```diff
@@ -14,6 +14,7 @@ export class DataItemSelect extends Component {
         className: PropTypes.string,
         label: PropTypes.string,
         value: PropTypes.string,
+        allowNone: PropTypes.bool,
```
amcgee: When would we not allow None?

turban: For thematic layers: 

skjermbilde 2018-11-21 kl 13 20 09

PR #63, opened 2018-11-23
Upgrade to d2-ui 5.0.10, use new dialog sizing prop

amcgee:
* Upgrade to d2-ui 5.0.10 which brings Interpretation dialog style fixes, smaller FileMenu button fontsize, and other improvements.
* There are now three statically-defined Dialog sizes - small, medium, and large - so remove custom Dialog sizing and use the maxWidth property to specify the size for each dialog.
* Decrease menubar font-size to 15px to match DV and d2-ui FileMenu button.


amcgee:
They look fine to me :-)

PR #65, opened 2018-11-27
Add global React ErrorBoundary

amcgee:
https://jira.dhis2.org/browse/DHIS2-5337

I tried to keep dependencies to a minimum so I'm not using theme variables.  It does depend on `withStyles` from Mui as well as a Material infoIcon.  Because of the Mui3 dependency I'm not going to add it to d2-ui-core at this time and will duplicate the file in Data-Visualizer (@jenniferarnesen for confirmation on that)


turban:
PR looks fine - but I think it would be better to include the DHIS2 logo, and maybe the app name on the  error page - so it looks like a error from the app and not from the browser. @joedhis 

The user should also be allowed to send a "crash report".

src/components/errors/FatalErrorBoundary.js
```diff
@@ -0,0 +1,81 @@
...
+
+    render() {
+        const { classes, children } = this.props;
+        if (this.state.hasError) {
+            return (
+                <div className={classes.mask}>
+                    <div className={classes.container}>
+                        <InfoIcon className={classes.icon} />
+                        <div className={classes.message}>
+                            Something went wrong
```
jenniferarnesen: Add Internationalization here and line 65

amcgee: Hmm thanks @jenniferarnesen good point, I think that's safe enough to include in the error boundary...?

amcgee:
> PR looks fine - but I think it would be better to include the DHIS2 logo, and maybe the app name on the error page - so it looks like a error from the app and not from the browser. @joeDHIS
> 
> The user should also be allowed to send a "crash report".

Thanks @turban - I opened [DHIS2-5400](https://jira.dhis2.org/browse/DHIS2-5400) to track the crash report issue.

I think a simple first step we could do here is to have an initially-hidden "Technical Details" panel which exposes details about the exception and could be copy-and-pasted in tech support cases.  I think that's minimally-intrusive but helpful when it is needed.  (@edoardo came up with this idea, I think it's would be a great addition here, @joeDHIS any feedback?)

I would LOVE to have an automatic feedback mechanism at some point, but that's probably a longer-term discussion (in [DHIS2-5400](https://jira.dhis2.org/browse/DHIS2-5400))

src/components/errors/FatalErrorBoundary.js
amcgee: Added i18n

amcgee:
@turban This is ready for review, Joe has stamped his approval on the design

amcgee:
Question @jenniferarnesen - should the "An error occurred in the DHIS2 Maps application" be translated?  It currently is, but maybe it should use variable substitution for `{{AppName}}` so translations could be shared between apps?  Don't know if we do that.

PR #66, opened 2018-11-27
Correct event colors from older map favorites

turban:
We now expect all colors to follow the CSS standard. Previously we stored colors without the #. This PR corrects these colors for events layers where we have this problem: https://jira.dhis2.org/browse/DHIS2-5371

The utility file colorscale.js was renamed to colors.js to make it more generic for color handling.

### DIFF
--- a/src/util/colors.js
+++ b/src/util/colors.js
@@ -1,3 +1,4 @@
+import { isString } from 'lodash/fp';
 import colorbrewer from '../constants/colorbrewer';
 
 // Allowed color scales from ColorBrewer for EE (needs to have at least 9 classes)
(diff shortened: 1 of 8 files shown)


src/util/colors.js
```diff
@@ -54,3 +55,11 @@ export const defaultColorScale = getColorPalette(
     defaultColorScaleName,
     defaultClasses
 );
+
+// Correct colors not adhering to the css standard (add missing #)
+export const cssColor = color => {
+    if (!isString(color)) {
```
amcgee: When will the color not be a string?

amcgee:
Looks good, added one comment

src/util/colors.js
```diff
@@ -54,3 +55,11 @@ export const defaultColorScale = getColorPalette(
...
+        return color;
+    }
+    return (/(^[0-9A-F]{6}$)|(^[0-9A-F]{3}$)/i.test(color) ? '#' : '') + color;
```
amcgee: This could still be an invalid color (like "blorange") but I don't think we want to do a full CSS spec color comparison so that's fine :-)

src/util/colors.js
turban: In case it is null or undefined.

src/util/colors.js
turban: Yes, that might happen - but these don't exist in the system at this point :-)

amcgee: 👍

src/util/colors.js
amcgee: 👍

PR #67, opened 2018-11-27
Show data element operand when loading a saved map

turban:
Issue: https://jira.dhis2.org/browse/DHIS2-5366

### DIFF
--- a/src/components/dataElement/DataElementOperandSelect.js
+++ b/src/components/dataElement/DataElementOperandSelect.js
@@ -44,11 +44,15 @@ export class DataElementOperandSelect extends Component {
             style,
         } = this.props;
 
-        if (!dataElementGroup) {
-            return null;
-        }
+        let items;
 
-        const items = dataElementOperands[dataElementGroup.id];
+        if (!dataElementOperands && !dataElementGroup && !dataElement) {
+            return null;
+        } else if (!dataElementGroup && dataElement) {
+            items = [dataElement];
+        } else {
+            items = dataElementOperands[dataElementGroup.id];
+        }
 
         return (
             <SelectField
--- a/src/components/dataElement/DataElementOperandSelect.js
+++ b/src/components/dataElement/DataElementOperandSelect.js
@@ -46,12 +46,12 @@ export class DataElementOperandSelect extends Component {
 
         let items;
 
-        if (!dataElementOperands && !dataElementGroup && !dataElement) {
-            return null;
-        } else if (!dataElementGroup && dataElement) {
+        if (dataElementOperands && dataElementGroup) {
+            items = dataElementOperands[dataElementGroup.id];
+        } else if (dataElement) {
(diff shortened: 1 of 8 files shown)


src/components/dataElement/DataElementOperandSelect.js
```diff
@@ -44,12 +44,16 @@ export class DataElementOperandSelect extends Component {
...
 
-        if (!dataElementGroup) {
+        let items;
+
+        if (!dataElementOperands && !dataElementGroup && !dataElement) {
             return null;
+        } else if (!dataElementGroup && dataElement) {
+            items = [dataElement];
+        } else {
+            items = dataElementOperands[dataElementGroup.id];
```
amcgee: This might be a little easier to follow as:

```js
if (dataElementOperands && dataElementGroup) {
  items = dataElementOperands[dataElementGroup.id];
} else if (dataElement) {
  items = [dataElement];
} else {
  return null;
}
```

Note that as written int the PR there as a possible bug if `dataElementOperand = null`, `dataElement = null`, and `dataGroup = 'someGroup'` because it will perform `dataElementOperands['someGroup']` which is `null['someGroup']` and an error.

src/components/edit/thematic/ThematicDialog.js
```diff
@@ -259,40 +259,41 @@ export class ThematicDialog extends Component {
...
-                                        key="totals"
-                                        operand={operand}
-                                        onChange={setOperand}
-                                        style={styles.select}
-                                    />
-                                ),
-                                operand === true ? (
-                                    <DataElementOperandSelect
-                                        key="element"
+                            {valueType === 'de' ||
```
amcgee: `valueType === 'de' || ...` is unnecessary here since we're immediately checking `valueType === 'dc'` right afterwards

src/components/edit/thematic/ThematicDialog.js
```diff
@@ -259,40 +259,41 @@ export class ThematicDialog extends Component {
...
+                                    />,
+                                    dataElementGroup && (
+                                        <TotalsDetailsSelect
+                                            key="totals"
+                                            operand={operand}
+                                            onChange={setOperand}
+                                            style={styles.select}
+                                        />
+                                    ),
+                                    operand === true || valueType === 'dc' ? (
```
amcgee: I believe `valueType` will ALWAYS `=== 'dc'` here since we're in the && condition of `valueType === 'dc'` above

src/components/dataElement/DataElementOperandSelect.js
turban: Yes, thats's better. Fixed.

src/components/edit/thematic/ThematicDialog.js
turban: valueType can be 'de' or 'dc' - see comment above.

PR #68, opened 2018-11-28
Fix issues with filter and style by data element

turban:
This PR fixes various issues with event layers filters and styling by data element:
- Added right padding to event legend 
- Allow dimensions to have empty items array if it contains a filter (issue with saved maps)
- Fix issue with header names needed for legend
- Don't include styleDataItem as a dimension if it's already included in a filter

Related issue: https://jira.dhis2.org/browse/DHIS2-5397

Sorry that the tests got reformatted when I edited them :-)

### DIFF
--- a/src/loaders/eventLoader.js
+++ b/src/loaders/eventLoader.js
@@ -172,6 +172,7 @@ export const loadData = async (request, config = {}) => {
     const response = await d2.analytics.events.getQuery(request);
 
     const { data, names } = createEventFeatures(response, config);
+
     return {
         data,
         names,
@@ -202,7 +203,7 @@ const getFilterOptionNames = async (filters, headers) => {
         optionSets.map(id =>
             d2.models.optionSets
                 .get(id, {
-                    fields: 'options[code,displayName~rename(name)]',
+                    fields: 'options[id,code,displayName~rename(name)]',
                 })
                 .then(model => model.options)
                 .then(options =>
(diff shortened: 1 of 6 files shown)


src/loaders/eventLoader.js
```diff
@@ -225,7 +226,8 @@ const getFilterOptionNames = async (filters, headers) => {
 };
 
 // Empty filter sometimes returned for saved maps
-const isValidDimension = ({ dimension, items }) =>
-    Boolean(dimension && (!items || items.length));
+// Dimension without filter and empty items array returns false
+const isValidDimension = ({ dimension, filter, items }) =>
+    Boolean(dimension && (filter || (!items || items.length)));
```
amcgee: Minor - The extra parenthesis shouldn't be required here - just `(filter || !items || items.length)`

src/util/geojson.js
```diff
@@ -86,17 +86,15 @@ export const buildEventCoordinateGetter = (headers, eventCoordinateField) => {
...
-        ...(config.outputIdScheme !== 'ID'
-            ? response.headers.reduce(
-                  (names, header) => ({
-                      ...names,
-                      [header.name]: header.column,
-                  }),
-                  {}
-              )
-            : null),
+        ...response.headers.reduce(
```
amcgee: Looking at this again I think we can actually leave out the `...` object spread operator now - it should just be:

```
const names = response.headers.reduce(
  ...
);

PR #71, opened 2018-11-28
Add loading text to layer card

turban:
Add "Layer is loading" message to the layer card, and avoids displaying the layer id as a name.


amcgee:
Looks good!  Just maybe have @joeDHIS take a quick look at the wording?

@joedhis the title is "Layer is loading" until we get a real name, is that good with you?

joeDHIS:
I think "Loading layer..." is best for the text label.

PR #76, opened 2018-12-04
Fix for style issues in download maps

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-5478

dom-to-image is not loading the Robot font for some reason. A quickfix was to switch to Arial when going into download mode. 

Added removeLineBreaks as a utility method. Used to remove line breaks from html text strings. 

map-th2ly

### DIFF
--- a/src/util/helpers.js
+++ b/src/util/helpers.js
@@ -139,4 +139,4 @@ export const getStartEndDateError = (startDate, endDate) => {
 
 // Remove line breaks from text (not displayed corrently in map downloads)
 // https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string/10805292#10805292
-export const removeLineBreaks = text => text.replace(/\r?\n|\r/g, '');
+export const removeLineBreaks = text => text.replace(/\r?\n|\r/g, ' ');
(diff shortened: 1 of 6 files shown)


src/util/helpers.js
```diff
@@ -136,3 +136,6 @@ export const getStartEndDateError = (startDate, endDate) => {
     }
     return null;
 };
+
+// Remove line breaks from text (not displayed corrently in map downloads)
+export const removeLineBreaks = text => text.replace(/(\r\n\t|\n|\r\t)/gm, '');
```
amcgee: We are we matching tabs here?

turban: Changed to /\r?\n|\r/g and added a comment about the source :-)

src/util/helpers.js
```diff
@@ -136,3 +136,7 @@ export const getStartEndDateError = (startDate, endDate) => {
     }
     return null;
 };
+
+// Remove line breaks from text (not displayed corrently in map downloads)
+// https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string/10805292#10805292
+export const removeLineBreaks = text => text.replace(/\r?\n|\r/g, '');
```
amcgee: Better, nice.  As that StackOverflow answer states though do we want to replace with a space instead of an empty string?  Is it possible there will be newlines between words?

turban: Should only be newlines between tags, but added a space in any case:

PR #77, opened 2018-12-05
Show relative period data from interpretations

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-5477

This PR displays the time of the selected interpretation on the map:

skjermbilde 2018-12-05 kl 10 25 46

The PR also includes a fix for map legends on dashboard maps: the layers are reorders so the top layer is shown first. 

We don't support different date formats/locales in the Maps app. I've created a separate issue for this task: https://jira.dhis2.org/browse/DHIS2-5498

For later: 
- Shared repo for app icons

### DIFF
--- a/src/components/map/MapName.js
+++ b/src/components/map/MapName.js
@@ -1,28 +1,69 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
+import i18n from '@dhis2/d2-i18n';
+import InterpretationIcon from '../core/InterpretationIcon';
+import { timeFormat } from 'd3-time-format';
+
+const formatTime = timeFormat('%b %d, %Y'); // TODO: Support different locales
 
 const styles = theme => ({
-    name: {
+    root: {
+        display: 'flex',
+        justifyContent: 'center',
         position: 'absolute',
-        top: theme.spacing.unit,
-        left: '50%',
-        transform: 'translateX(-50%)',
+        maxWidth: 'calc(100% - 100px)',
+        top: 0,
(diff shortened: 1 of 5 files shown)


src/components/map/MapName.js
```diff
@@ -1,28 +1,69 @@
...
-        padding: '6px 8px',
-        backgroundColor: 'rgba(255,255,255,0.9)',
-        boxShadow: theme.shadows[1],
-        borderRadius: theme.shape.borderRadius,
         fontSize: theme.typography.fontSize,
+        '& > div': {
+            backgroundColor: 'rgba(255,255,255,0.9)',
+            boxShadow: theme.shadows[1],
+            borderRadius: theme.shape.borderRadius,
+            margin: 8,
```
amcgee: Nitpick - can we use `theme.spacing.unit` here?

src/components/map/MapName.js
```diff
@@ -1,28 +1,69 @@
...
+        },
+    },
+    name: {
+        padding: '6px 8px',
+        lineHeight: '17px',
+    },
+    interpretation: {
+        position: 'relative',
+        padding: '7px 8px 7px 28px',
```
amcgee: 7 is an odd number, is there a reason to use 7 instead of `theme.spacing.unit === 8` ?

turban: The font is smaller for the interpretation note so it was to get the styling right :-) Can't use height as it might span more than one line.

src/components/map/MapName.js
turban: Fixed.

PR #79, opened 2018-12-05
Add program id to tracked entity attributes

turban:
Fixes issue: https://jira.dhis2.org/browse/DHIS2-5457

@larshelge, are Program Tracked Entity Attributes supposed to work as an event data item for thematic layers? 

I saw that we get an error if we select an tracked entitiy attribute in the Maps app:

skjermbilde 2018-12-05 kl 15 35 28

https://play.dhis2.org/2.31/api/31/analytics.json?dimension=ou:LEVEL-2&dimension=dx:lw1SqmMlnfh&filter=pe:2018&displayProperty=SHORTNAME

Returns error: "Dimension dx is present in query without any valid dimension options"

The reason is that the program id is not included in the dx dimension as for program data elements. This PR adds it: 

https://play.dhis2.org/2.31/api/31/analytics.json?dimension=ou:LEVEL-2&dimension=dx:uy2gU8kT1jF.lw1SqmMlnfh&filter=pe:2018&displayProperty=SHORTNAME

It avoids the error, but I get no data (rows) back.

skjermbilde 2018-12-05 kl 15 29 09

Is this meant to work and is the syntax differently for  tracked entitiy attributes?

PS! I cheked the old GIS app, and selecting a tracked entitiy attribute don't to work there either, as no api request is generated.

### DIFF
--- a/src/components/dataItem/EventDataItemSelect.js
+++ b/src/components/dataItem/EventDataItemSelect.js
@@ -77,7 +77,13 @@ export class EventDataItemSelect extends Component {
             dataElements[program.id],
             null,
             excludeValueTypes
-        );
+        ).map(item => ({
+            ...item,
+            id:
+                item.id.indexOf('.') === -1
+                    ? `${program.id}.${item.id}`
+                    : item.id, // Add program id to tracked entity attributes
+        }));
 
         return (
             <SelectField


larshelge:
Yes this is the correct syntax.

<program-id>.<dataelement-id>

I need to look closer if there is a potential API issue, however the syntax is correct so please go ahead and merge the fix.

PR #83, opened 2019-01-09
Linting fixes

turban:
This PR fixes all linting errors in the source code, and adds a lint check to the pre-commit hook.


src/components/map/Layer.js
```diff
@@ -12,13 +12,16 @@ class Layer extends PureComponent {
...
+        data: PropTypes.array,
+        dataFilters: PropTypes.array,
         id: PropTypes.string.isRequired,
         index: PropTypes.number,
         editCounter: PropTypes.number,
         opacity: PropTypes.number,
         isVisible: PropTypes.bool,
         config: PropTypes.object,
-        // labels: PropTypes.object, // TODO: Is sometimes boolean
+        labels: PropTypes.bool,
```
amcgee: Is this always bool or should it be a union type?

src/loaders/thematicLoader.js
```diff
@@ -70,7 +70,7 @@ const thematicLoader = async config => {
         title: name,
         period: period
             ? names[data.metaData.dimensions.pe[0]]
-            : `${formatLocaleDatee(config.startDate)} - ${formatLocaleDate(
+            : `${formatLocaleDate(config.startDate)} - ${formatLocaleDate(
```
amcgee: YAY LINTING!

src/reducers/map.js
```diff
@@ -260,10 +265,12 @@ const map = (state = defaultState, action) => {
                 ...state,
             };
 
+        /*
```
amcgee: Should we just delete this case?

src/util/sld.js
```diff
@@ -25,10 +25,12 @@ export const createSld = () => {
 
         let stopFilter = 'PropertyIsLessThan';
 
+        /*
```
amcgee: Any reason to keep this commented block?  Why was is safe to remove?

src/loaders/thematicLoader.js
turban: Need to backport this fix to 2.31. Will do.

src/components/map/Layer.js
amcgee: What was the resolution here?

turban: It is always boolean

src/loaders/thematicLoader.js
turban: Fix backported to 2.31

PR #85, opened 2019-01-17
Smoke test plugin.html

amcgee:
Use Cypress to test basic functionality of plugin.html.  This should ideally be extended to better test true functionality.


amcgee:
The Travis build failed because I had forgotten to merge the fix from master, so it correctly caught the regenerator-runtime failure in `plugin.html` :-D

PR #86, opened 2019-01-18
feat: dynamic dimension filtering for thematic layers

turban:
Solves: https://jira.dhis2.org/browse/DHIS2-301

This PR upgrades React to 16.8.6 and Material UI to 3.9.3, which was required by DimensionsPanel in d2-ui-analytics. 

The backend is not yet supporting multiple data dimensions as filters, but it according to @larshelge it will be fixed before 2.32 is released: https://jira.dhis2.org/browse/DHIS2-6544

I don't think we should wait for DHIS2-6544 before we merge this PR into master, as we need people to test it before launch. 

The following data dimensions are supported:
- CATEGORY
- CATEGORY_OPTION_GROUP_SET
- ORGANISATION_UNIT_GROUP_SET

A new filter tab is added to thematic layers with simal design as event filters. The applied filters are displayed in the legend. 

No units tests are currently added, needs to be done after code freeze.


amcgee:
Code looks good and filters work!  Found one issue when selecting "pregnant / not pregnant" as a filter.  I think this is OK to merge and fix the issues after tonight though.

Screenshot 2019-04-09 19 49 19

PR #90, opened 2019-02-05
Support different map engines

turban:
Make sure you have @dhis2/gis-api@^32.0.1 or higher in dependencies before testing this PR.

This PR adds support for multiple map engines by replacing all Leaflet specific map methods and event handling. 

- [x] Added "hasLayerSupport" check if basemap is supported. Used to exclude Google basemaps.
- [x] Allow opacity and visibility to be specified when the layer is created. 
- [x] Move layer order handling from the Maps app to the map engine. 
- [x] Use index property to specify layer order.
- [x] Added "onClick" and "onRightClick" callbacks for all layers. The response should be the same from all map engines. 
- [x] Move popup handling from the Maps app to the map engine. 
- [x] Pass access token as a promse to Google Earth Engine layers.
- [x] Move buffer handling from the Maps app to the map engine (part of layer config).
- [x] Switch to control position system used by Mapbox GL JS.
- [x] Use new "resize" engine method instead of "invalidateSize".
- [x] Switch to coordinate order used by Mapbox GL JS (bounds and [lng, lat]).
- [x] Code cleaning.


src/components/map/MapProvider.js
```diff
@@ -1,6 +1,8 @@
 import { Component } from 'react';
 import PropTypes from 'prop-types';
-import d2map from '@dhis2/gis-api';
+// import D2map from '@dhis2/gis-api';
+import D2map from '@dhis2/gis-api/src';
```
amcgee: This should use the `import D2map from '@dhis2/gis-api'` form, which works if you build `gis-api` before linking it

amcgee: See the steps I took to link (which may be incorrect?  In [gis-api#45](https://github.com/dhis2/gis-api/pull/45)

turban:
Need to move on, so I'm merging this PR into master. Will do some extra testing of the app in the coming days.

PR #92, opened 2019-02-18
chore: update cypress dependency

amcgee:
Update Cypress dependency to fix an issue with end-to-end tests.

CC @jenniferarnesen, might need to do this in DV also


amcgee:
The previous behavior this appears to fix was (from @turban):

skjermbilde 2019-02-18 kl 13 21 36

PR #94, opened 2019-02-19
Open chart as map / open map as chart

turban:
This PR allows the user to open a chart in The Data Visualizer as a map, and vice versa. The analytical object is kept in the user data store when the app is switched.

It the analytical object contains multiple data dimensions we open a dialog to allow the user to select between them. One thematic layer will be created per data dimension. 

This PR also adds new relative periods, to keep them synced with Data Visualizer: https://jira.dhis2.org/browse/DHIS2-6066

How to test locally:
- clone the Data visualizer repo: https://github.com/dhis2/data-visualizer-app (master branch)
- yarn install / yarn build / yarn start-app
- create/open a chart 
- open the chart type panel and select "Open as: Map"
- the analytical object should now be stored in the user data store
- open the Maps app with "currentAnalyticalObject": 
http://localhost:8082/?currentAnalyticalObject=true
- test going from Maps app to Data visualizer app by clicking "Open as chart" for a thematic layer. 

Notes:
- Going from a map to a (column) chart is more straightforward than the other way, as this should always work. 
- When we go from a chart to a map we often need to move the dimensions between to rows/columns/filters, and this could result in issues. This should work most of the time :-)
- There are currently limitations when having relative periods as filters, and this could result in strange results - this should be fixed in the backend. 
- If we later add support for multiple period for maps (https://jira.dhis2.org/browse/DHIS2-2255), we could support more chart layouts. 
- Maps app should use the same routing pattern as Data Visualizer (separate issue).
- We should improve error handling across the app (separate issues).

skjermbilde 2019-03-07 kl 14 37 18

screen shot 2019-03-02 at 15 58 45


amcgee:
Hi @turban is this PR ready for me to take a look at or is it still a WIP?

amcgee:
@turban I looked shallowly through the code changes and they look good!  In testing I ran into this one issue:

When going from `ANC: 1-4 visits last 12 months` saved AO in DV (http://localhost:3000/#/Tun9tJb3sQt) and selection `Open as Map`, then navigating to http://localhost:8082/?currentAnalyticalObject=true I correctly get the popup to select layers to add.  This looks great!  However when adding the layers I get an error saying the orgUnit (which is the root Sierra Leone) doesn't exist:

screen shot 2019-03-08 at 1 46 38 am

Selecting a different Org Unit (`Bo`) in DV before switching works as expected.

Second issue - when editing the added layer the Indicator field is empty.

Also a side note (CC @neeilya @JoeDHIS) should the app clear the currentAnalyticalObject from the Data Store after successfully switching?  Right now if you, say, have a tab open with "maps?currentAnalyticalObject=true" or "data-visualizer#/currentAnalyticalObject" and you make changes to the chart or map, then you reload the page you will go back and load whatever the last "openAs" chart or map was, even if it was something you did in another tab.  Similarly, if you click "cancel" in the dialog presented when opening a chart as a map you still have `currentAnalyticalObject=true` in the queryString

turban:
Thanks for your review @amcgee! 

First issue: This is expected behaviour as we don't have any coordinates (polygon) for the root org unit in the demo db. I think we should add root org unit polygon, but it's a separate issue. 

Second issue: I'm getting the selected data element, but the data element group is empty. This is by design as we don't have information about the group in the analytical object (same when you open a saved map). 
screen shot 2019-03-08 at 19 38 31

Did you see an empty indicator field for another analytical object?

@neeilya, what do you do in the DV app - do you clear the analytical object from the user data store after the chart is rendered? 

I have pushed a fix that clears the "currentAnalyticalObject=true" from the URL after the analytical object is read.

amcgee:
> First issue: This is expected behaviour as we don't have any coordinates (polygon) for the root org unit in the demo db. I think we should add root org unit polygon, but it's a separate issue.

Thanks for the clarification, we should get that cleared up in the demo DB but it would be good also to look into making DHIS2 more resilient to those kinds of misconfigurations.  I assume this means that you can "open chart as map" and get a map that it is not possible to create directly in the Maps app itself?

> Second issue: I'm getting the selected data element, but the data element group is empty. This is by design as we don't have information about the group in the analytical object (same when you open a saved map).

This seems wrong to me... how can we not know the group if we know the data element (which belongs to exactly one group)?  Do we need to pass more information from DV or make a clarifying request to the server when we receive an AO without a Group?  If a map layer is created it should be editable and reproducible from the Layer Add/Edit dialog right?

@neeilya can you comment on the final point?

These may be issues to be dealt with separately from this PR (which looks good), but worth discussing.

turban:
> These may be issues to be dealt with separately from this PR (which looks good), but worth discussing.

I agree @amcgee, and I've created two new issues to track them: 
https://jira.dhis2.org/browse/DHIS2-6228
https://jira.dhis2.org/browse/DHIS2-6229

neeilya:
@turban 

>@neeilya, what do you do in the DV app - do you clear the analytical object from the user data store after the chart is rendered?

No, my initial thought was that manually removing it was redundant since opening chart in another app rewrites the chart config in user data store.

neeilya:
>This seems wrong to me... how can we not know the group if we know the data element (which belongs to exactly one group)? Do we need to pass more information from DV or make a clarifying request to the server when we receive an AO without a Group? If a map layer is created it should be editable and reproducible from the Layer Add/Edit dialog right?
>
>@neeilya can you comment on the final point?

@amcgee this is specific for maps-app I think, cause in DV we can have multiple data elements for one chart, so there is no data element group field for saved AO/chart.

turban:
I'm merging this into master now, if there are other issues we'll fix it in a separate PR.

PR #99, opened 2019-03-01
Update/d2 ui interpretations panel

joakimia:
This PR includes new interpretation panel with rich text, textfield in the panel instead of a dialog and styling changes like icons with tooltips on hover according to the [design document](https://projects.invisionapp.com/share/A7LT4TJYETS#/screens/339941736_Specs-Comment).


package.json
```diff
@@ -48,7 +48,7 @@
         "@dhis2/d2-i18n": "^1.0.4",
         "@dhis2/d2-ui-core": "5.1.3",
         "@dhis2/d2-ui-file-menu": "5.1.3",
-        "@dhis2/d2-ui-interpretations": "5.1.3",
+        "@dhis2/d2-ui-interpretations": "^5.2.8",
```
amcgee: Please update all d2-ui components, since they all have the same version, unless there's a particular reason not to.  Also we're using the explicit version pattern for `d2-ui` dependencies here, so let's stick with that unless we want to change all of them.

src/components/interpretations/InterpretationsPanel.js
```diff
@@ -68,15 +68,17 @@ class InterpretationsPanel extends Component {
...
-                <Interpretations
-                    d2={this.context.d2}
-                    id={mapId}
-                    type="map"
-                    currentInterpretationId={interpretationId}
-                    onCurrentInterpretationChange={
-                        this.onCurrentInterpretationChange
-                    }
-                />
+                <div>
```
amcgee: Why is this extra DOM node required?

package.json
joakimia: done! updated all `d2-ui` using explicit version.

src/components/interpretations/InterpretationsPanel.js
joakimia: Without the wrapper DOM, the Drawer does not produce a scrollbar after changing the root interpretation component div to be flex container. I had to wrap it inside an extra div in order for it to overflow with a scrollbar. I've tried to add/modify the Drawer's css rules but the scrollbar does not appear when the component overflows vertically.

amcgee: Hmm... would it be possible to move that wrapper div into the `<Interpretations>` component then?  It's not obvious to a consumer of the component that they need to wrap it in a `<div>`.  If it's too much effort to rework the component now let's add a GitHub or JIRA issue to make sure we clean it up later.

package.json
amcgee: Thanks @joakimia but it looks like you're using `^` version ranges (`^5.2.8`) rather than explicit versions (`5.2.8`).  Minor detail but since that's what we were using here before (and we're not necessarily great at sticking to `semver` guarantees yet) let's stick with it.

src/components/interpretations/InterpretationsPanel.js
joakimia: Yeah, I moved the wrapper to the component itself and updated the dependencies 👍

package.json
joakimia: I've updated all d2-ui packages to use explicit version after moving the div wrapper 👍

amcgee:
@joakimia before we merge this can you merge this with master and make sure the end-to-end tests pass in Cypress?

amcgee:
(They didn't run on your previous commits because #100 re-enabled them)

joakimia:
@amcgee Merged with master, the cypress tests are failing:

` CypressError: Timed out retrying: Expected to find element: 'header', but never found it. `

Tried to update the header bar just to see if there were any changes, but the same error messages are received. 

Reverted that commit.

Birkbjo:
Tests failed due to userGroup being included when initializing d2. Re-ran cy:e2e:update to include this request as well.

joakimia:
Sat down with Birk and found that first cypress received a 404 when trying to fetch `userSettings`. Somehow, this error just dissipated during the investigation (Switched between testing his [PR](https://github.com/dhis2/maps-app/pull/102), master and my PR locally - which all received 404, but dissipated magically).

After the userSetting stub request succeeded, we updated the fixture with userGroup which solved the travis build issue. Netlify were struggling with initializing,  the deploy preview shows that it was unable to fetch userSettings from dev, but the instance was down.

```
3:20:31 PM: Executing user command: yarn install && yarn build && yarn run manifest package.json build/manifest.webapp --manifest.activities.dhis.href="https://play.dhis2.org/dev"
```

Possible to trigger a re-deploy?

PR #100, opened 2019-03-05
Fix Cypress tests on Travis

amcgee:
That was a battle!

This is not the ideal solution to this problem, but at least it allows us to re-activate Cypress tests on Travis.  1 million thanks to @varl and @Birkbjo for the midnight oil trying to track this bad boy down.  Does 114 messages set the record for a Slack thread?

In the end it was absolute URLs in API payloads that caused this obscure RXJS error: `Uncaught TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.` for all but the basic `loads` smoke test in the Cypress end-to-end test runner with stubbed network.

The issue occurred because:

1. The API returns an `apiEndpoint` with an absolute URL (i.e. `https://play.dhis2.org/dev/...`) for many endpoints.  This causes issues when trying to use that response payload with a different server or application configuration (even if it's the same server accessed through a proxy).  Ideally, the API should be idempotent (a related but different issue) and use *relative links* rather than absolute ones.
2. The infrastructure for stubbing network requests built as a Cypress plugin masks out the basepath for API request URLs but doesn't mutate the response payload if it includes references to the basepath used when recording.  Because of this, stubbing the network against a different server URL works until the app (and in particular [d2](https://github.com/dhis2/d2)) dutifully follows an absolute link for a subsequent request and can't find it (because that server isn't being stubbed).  Because I recorded updated fixtures on my machine which uses a non-standard DHIS2 server url (`d2-cluster.winged.tech`) the stubbed Cypress tests worked fine for me locally but failed miserably when run in Travis (which pointed the app at `localhost:8080`) by default.
3. The maps app should be resilient to network failures, but it is not.  In this case a network 404 was passed through directly as an invalid object to a `redux-observable`/`rxjs` subscription function, and the error bubbled all the way up so that it had to be caught by Cypress.

This change simply replaces `d2-cluster.winged.tech` with `localhost:8080` in the network fixture files, which fixes the build on Travis and allows us to re-enable Cypress tests.  However it *DOES NOT SOLVE ANY OF THE ROOT CAUSES*.  To do that (for each cause above) we need to:

1. **DIFFICULTY: HARD?** Stop sending absolute URLs in API responses and stop following absolute URLs in `d2` (this could also possibly be exploited to leak system information to 3rd party URLs in very rare edge cases?)
2. **DIFFICULTY: MEDIUM** Wrap the test script to stabilizes and synchronizes the API URL used for the running app and the Cypress test-runner.  Make sure that the server used to record network requests is always running with the same base url (until #1 is fixed)
3. **DIFFICULTY: MEDIUM** Seek and destroy all places where we are not handling network request failures in the app and `d2-ui` components.  Alternatively, build error-handling into the API request layer (currently `d2`)

Any other suggestions to make sure we don't run into this again!?


varl:
>**DIFFICULTY: HARD?** Stop sending absolute URLs in API responses and stop following absolute URLs in `d2` (this could also possibly be exploited to leak system information to 3rd party URLs in very rare edge cases?)

So. I did some digging in the core, and there is a way to get the relative API urls:

We are using this: https://play.dhis2.org/dev/api/31/schemas/dataElement.json?fields=[apiEndpoint,name] which returns:

```
{
  "apiEndpoint": "https://play.dhis2.org/dev/api/dataElements",
  "name": "dataElement"
}
```

We should probably use: https://play.dhis2.org/dev/api/31/schemas/dataElement.json?fields=[relativeApiEndpoint,name] which gives us:

```
{
  "relativeApiEndpoint": "/dataElements",
  "name": "dataElement"
}
```

These are the files I've found which uses `apiEndpoint`:
   
Libs:                 

- ./d2/lib/d2-browser.js                                                     
- ./d2/lib/d2-browser.min.js                                                                                    
- ./d2/src/d2.js                                                             
- ./d2/src/model/ModelCollectionProperty.js                                  
- ./d2/src/model/ModelDefinition.js                                                        
- ./d2/src/model/config/index.js                                             
- ./d2/src/model/helpers/models.js                                            
- ./d2-ui/packages/translation-dialog/src/translationForm.actions.js                            
- ./ui/src/utils/api.js        

Apps:
               
- ./charts-app/src/index.js                                                    
- ./charts-app/src/plugin.js                                                       
- ./data-visualizer-app/packages/app/src/api/visualization.js                      
- ./event-charts-app/src/index.js                                              
- ./event-charts-app/src/plugin.js                                             
- ./event-reports-app/src/index.js                                             
- ./event-reports-app/src/plugin.js                                                   
- ./maintenance-app/src/EditModel/event-program/notifications/epics.js         
- ./maintenance-app/src/EditModel/option-set/actions.js                        
- ./maintenance-app/src/config/custom-models/locale/localeSchemaDefinition.js                                                   
- ./pivot-tables-app/src/index.js                                              
- ./pivot-tables-app/src/plugin.js                                                                  
- ./translations-app/src/pages/Translations/index.js                           
- ./translations-app/src/utils/testFaker.js                                    

Some of these might be false positives (find/grep search), but even so I think it's doable to do this.

Are there any reasons we shouldn't use the `relativeApiEndpoint`?

>**DIFFICULTY: MEDIUM** Wrap the test script to stabilizes and synchronizes the API URL used for the running app and the Cypress test-runner. Make sure that the server used to record network requests is always running with the same base url (until #1 is fixed)

I don't think we need this if we can patch our usage of `apiEndpoint`s to `relativeApiEndpoints`.

>**DIFFICULTY: MEDIUM** Seek and destroy all places where we are not handling network request failures in the app and `d2-ui` components. Alternatively, build error-handling into the API request layer (currently `d2`)

Even if we change `apiEndpoint` to `relativeApiEndpoint`, it is bad behavior for a library to not handle errors. Those bugs that occur inside of dependencies are a pain to every consumer of that library. We should _absolutely_ patch the API request layers to handle errors in a traceable way.

amcgee:
> So. I did some digging in the core

Awesome 🎉 , Thanks for digging!  That's great that a `relativeApiEndpoint` exists, we should prefer it religiously.  For reference across all the requests for all the e2e tests in `maps` I replaced 640ish absolute URLs.

What I'd like to understand better (more digging) is how and why `d2` follows the absolute URLs.  I believe it has to do with the initial download of schemas, but it is not consistent (the nominal behavior for APIs seems to be to use the BaseURL passed to d2 rather than the absolute one returned from the API).  We'll need to understand that better to make sure nothing breaks when switching to the `relativeApiEndpoint` field.

> Even if we change apiEndpoint to relativeApiEndpoint, it is bad behavior for a library to not handle errors.

I agree **100%**!  The API request layer should handle the errors, but there will also need to be some handling at the app layer as well to gracefully acknowledge and respond to the failure caught at the API layer.

> I don't think we need this [consistent tooling around  if we can patch our usage of `apiEndpoint`s to `relativeApiEndpoints`.

True, but standardizing that would be a nice additional guarantee of repeatable test environments.

@varl since all of these require a bit more work across multiple repos, should we merge this PR to re-enable end-to-end testing in maps for now and create a separate issue for DRY api behavior?

varl:
> We'll need to understand that better to make sure nothing breaks when switching to the `relativeApiEndpoint` field.

>The API request layer should handle the errors, but there will also need to be some handling at the app layer as well to gracefully acknowledge and respond to the failure caught at the API layer.

>True, but standardizing that would be a nice additional guarantee of repeatable test environments.

Agreed x 3.

I talked to @mortenoh on Slack and the `relativeApiEndpoint` is used to construct the `apiEndpoint`, so it is safe to use (noting here as well for posterity).

Birkbjo:
> The API request layer should handle the errors, but there will also need to be some handling at the app layer as well to gracefully acknowledge and respond to the failure caught at the API layer.

I believe the error-handling code in this case is a one-off thing, as it was trying to do some more complex stuff (returning multiple actions) and one failed to realize that the error was not handled correctly.

Longer story: What happened was that a promise that returned an array was merge-mapped, and then concatAll()'ed, which means that the stream now contains an array of objects, instead of a promise. Which is fine since its an array, and when that is collected each element is emitted in sequence, which is the desired behavior. However, when it fails, the action returned in the catch handler is collected by the concatAll() and the pure action-object is returned, rather than an observable of the action.

When working with redux-observable you really have to thing twice if you start chaining operators directly on the action stream. 
So I believe the bug was basically a misunderstanding of rxjs. Which it seems we are going to migrate from,.

amcgee:
Yeah, thanks @Birkbjo for the clarification of where this specific error came from!  Do you think you could open a PR to address it here in maps?

I think what @varl and I are after is more built-in resiliency so that errors are safely handled by default in a user-friendly way.

Unfamiliarity with `rxjs` and Observable / stream patterns is a problem as well - I for one am not familiar enough with the API or best practices to easily analyze redux-observable logic.

PR #101, opened 2019-03-06
Empty and disable org unit levels and groups selections when user org unit is selected

turban:
This PR fixes issue: https://jira.dhis2.org/browse/DHIS2-6190

When selecting an user org unit level, all other org unit selections are cleared and disabled. This is fixed for thematic, facility and boundary layers. 

This is the same behaviour as the Data Visualizer app.

At a later stage we should switch to the same org unit selection components as the Data visualizer app. 

This fix needs to be backported to Map app 2.29.

### DIFF
--- a/src/components/edit/BoundaryDialog.js
+++ b/src/components/edit/BoundaryDialog.js
@@ -117,6 +117,7 @@ class BoundaryDialog extends Component {
 
         const orgUnits = getOrgUnitsFromRows(rows);
         const selectedUserOrgUnits = getUserOrgUnitsFromRows(rows);
+        const hasUserOrgUnits = !!selectedUserOrgUnits.length;
 
         return (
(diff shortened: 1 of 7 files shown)


joeDHIS:
> Tested, looks good to me!
> 
> @joeDHIS the behavior here and in data-vis is to **clear** the selected orgunit levels and groups when selecting a user orgunit. When de-selecting the user orgunit, the previous levels and groups aren't restored. I think this makes sense, but it could be jarring for the user so might be worth an improvement down the road.

Thanks for the heads up @amcgee, I agree that's not ideal to clear the data. I'll keep this in mind for the new global org. unit selector. (Which is being documented/discussed [here](https://jira.dhis2.org/browse/DHIS2-6181), if you're interested. Feedback always welcome)

PR #105, opened 2019-03-11
v31/update interpretaion panel and d2-ui-packages,

joakimia:
This PR includes updating version of d2-ui packages.


joakimia:
Changed the link placeholder from `link-title` to `link-url` and updated the version 👍

PR #112, opened 2019-04-01
Translate old map favorite object in maps plugin

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-6501

This PR will transform the old map favorite format (having thematic layers 1-4) to the current format, before loading/drawing the map layers.


turban:
> Just one thing, is `upgradeGisAppLayers` still needed in `util/requests`?

Currently yes, it's used when an old analytical object is loaded directly by the Maps app or the plugin.

PR #114, opened 2019-04-02
feat: event polygons

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-3164

Associated PR for the GIS API: https://github.com/dhis2/gis-api/pull/52

This PR adds support for event polygons (including support for polygons in client and server clusters). The polygons can be styled as buffered like points.


amcgee:
Generally looks good, I think cypress tests are breaking because they expect a different format in the redux store?  Not sure.

turban:
> Generally looks good, I think cypress tests are breaking because they expect a different format in the redux store? Not sure.

Any advice on how to fix? We're not running cypress tests on commit?

amcgee:
@turban No, it's not super fast so running on commit might be prohibitive.  I can look into the failure tonight, but it might be more than just cypress since Netlify is failing too.

turban:
I've merged https://github.com/dhis2/gis-api/pull/52 and updated this PR to use GIS API 32.0.18 with event polygons support.

PR #115, opened 2019-04-03
fix: bind function for tei click handling

amcgee:
This fixes an issue with click handling of entities in TEI layers.  I don't know why the arrow parameter version wasn't working, but `this.onEntityClick` was passing as `undefined` to the GeoJSON layer construction.


amcgee:
Oh, looks like you already did this @turban

PR #129, opened 2019-04-24
fix: upgrade to version 32 of the web api

turban:
This PR switches to version 32 of the Web API.


turban:
> To confirm, this was a regression in the API response right?

True

turban:
NB! I changed the purpose of this PR to only include the web api upgrade. The other change is now in a separate PR: https://github.com/dhis2/maps-app/pull/131

Cypress tests are currently not running.

amcgee:
@turban it looks like the Cypress tests pass on Travis, I will test locally here too

PR #130, opened 2019-04-24
feat: support saving and loading of TEI relationship maps

amcgee:
Fixes #130 
Fixes [DHIS2-6746](https://jira.dhis2.org/browse/DHIS2-6746)

Add support for saving and loading TEI layer relationship configuration through the JSON-encoded `config` property of the MapView


amcgee:
> Could it be cleaner to store the properties directly on the config object and not in a relationships object?

Maybe, but I like the idea of namespacing them separately from i.e. the ee properties.  I think it would be cleanest if the config object in memory was similarly namespaced and didn't require separate actions/reducers for each property, but that's a refactor for another time.

If you'd prefer a flat `config.config` JSON object I can flatten it, let me know.

amcgee:
👍

PR #133, opened 2019-04-25
Validity check of program stage data elements and tracked entity attributes

turban:
Fixes app crash: https://jira.dhis2.org/browse/DHIS2-6733

The root cause to this issue is that the Web API allows an empty data item (no properties) to pass through: https://jira.dhis2.org/browse/DHIS2-6760

With this PR we validate that all program stage data elements and tracked entity attributes contains id, name and valueType properties which are required by the app. 

This replaces the fix in https://github.com/dhis2/maps-app/pull/128 which is reverted im this PR.

### DIFF
--- a/src/util/helpers.js
+++ b/src/util/helpers.js
@@ -125,10 +125,12 @@ export const addOrgUnitPaths = mapViews =>
 // https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string/10805292#10805292
 export const removeLineBreaks = text => text.replace(/\r?\n|\r/g, ' ');
 
-const mandatoryProps = ['id', 'name', 'valueType'];
+const mandatoryDataItemAttributes = ['id', 'name', 'valueType'];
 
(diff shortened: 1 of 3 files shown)


src/util/helpers.js
```diff
@@ -123,3 +124,11 @@ export const addOrgUnitPaths = mapViews =>
 // Remove line breaks from text (not displayed corrently in map downloads)
 // https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string/10805292#10805292
 export const removeLineBreaks = text => text.replace(/\r?\n|\r/g, ' ');
+
+const mandatoryProps = ['id', 'name', 'valueType'];
```
amcgee: Minor comment - since this variable is global to the file and the file isn't specific to data items it's probably better to name the variable a bit more descriptively, like `mandatoryDataItemAttributes` - longer, but could be less confusing in the future

turban: Will fix and merge!

PR #134, opened 2019-04-25
fix: show an alert when a load error occurs

amcgee:
This catches any error that occurs while loading a layer and displays an alert.  In the special case of 403 and 409 (permissions) errors we show a special "Access denied" error.  The layer is still added, so it can be edited or removed manually.

I don't love the mutation of the config object but it is what it is for now.


amcgee:
Screenshot 2019-04-25 14 47 14

PR #137, opened 2019-05-13
fix: correctly look up lowercase option code

amcgee:
This fixes [DHIS2-6913](https://jira.dhis2.org/browse/DHIS2-6913) which was introduced by a change to use lowercase strings in the OptionSet map (a15cd4f6904ac163c953fa7517c51869f27e61d3)


amcgee:
@turban told me he approves this message, but two-factor auth prevents him from saying so here ;-)

PR #138, opened 2019-05-16
fix: expand the width of drop down menus

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-4726 for Maps 2.30

The fix allows drop down menus to have a width of its content, rather than the width of the select field that triggers it. 

I applied a quickfix for this backport as the original select field component is in another repo that is no longer maintained. I've clicked through all layer dialogs and did't notice any unwanted side effects. 

Skjermbilde 2019-05-16 kl  16 38 28
Skjermbilde 2019-05-16 kl  16 37 53

### DIFF
--- a/scss/layout.scss
+++ b/scss/layout.scss
@@ -12,4 +12,5 @@ body {
 /* Fix to allow menus to have a width that fits its content in Material-UI 0.20 */
 div[role='presentation'] {
     width: auto !important;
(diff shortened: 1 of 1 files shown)


turban:
@amcgee good point - added max-width: 400px - the component already includes styles for scroll if the content exceeds this.

PR #147, opened 2019-06-05
Period selection fix for EE layers

turban:
A small refactoring to show the year dropdown when the user is editing an EE layer which is already created (and collections are loaded). 

Related issue: https://jira.dhis2.org/browse/DHIS2-6321

### DIFF
--- a/src/components/earthengine/Collection.js
+++ b/src/components/earthengine/Collection.js
@@ -51,31 +51,39 @@ export class CollectionSelect extends Component {
 
         if (id && !collections[id]) {
             loadEarthEngineCollection(id);
+        } else {
+            this.setYears();
         }
     }
 
     componentDidUpdate(prevProps) {
-        const { id, filter, collections } = this.props;
+        const { id, collections } = this.props;
 
         if (id && collections[id] !== prevProps.collections[id]) {
-            const yearItems = collections[id]
-                .filter(item => item.year)
-                .map(item => item.year);
+            this.setYears();
+        }
+    }
 
-            if (yearItems.length) {
-                // Get unique years
-                const years = [...new Set(yearItems)].map(year => ({
-                    id: year,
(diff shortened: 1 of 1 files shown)


src/components/earthengine/Collection.js
```diff
@@ -51,31 +51,39 @@ export class CollectionSelect extends Component {
 
         if (id && !collections[id]) {
             loadEarthEngineCollection(id);
+        } else {
```
amcgee: For completeness I think this should be `else if (collections && collections[id])` to avoid a possible undefined object issue on line 70.

Can `collections` ever be null?  If not it should at least be required in `propTypes`.  If yes (or even if not) we should check it before de-referencing `collections[id]`

src/components/earthengine/Collection.js
```diff
@@ -51,31 +51,39 @@ export class CollectionSelect extends Component {
 
         if (id && !collections[id]) {
```
amcgee: possibly `id && (!collections || !collections[id])` ?

src/components/earthengine/Collection.js
```diff
@@ -51,31 +51,39 @@ export class CollectionSelect extends Component {
...
+        } else {
+            this.setYears();
         }
     }
 
     componentDidUpdate(prevProps) {
-        const { id, filter, collections } = this.props;
+        const { id, collections } = this.props;
 
         if (id && collections[id] !== prevProps.collections[id]) {
```
amcgee: For completeness should this be `if (id && collections && collections[id] && collections[id] !== prevProps.collections[id])` ?  Will `collections[id]` ever **become** null?

src/components/earthengine/Collection.js
turban: I made collections required - the start value is set to an empty object in the store.

PR #154, opened 2019-06-26
feat: split view maps

turban:
Big PR alert! 

Depends on https://github.com/dhis2/gis-api/pull/58
JIRA-issue: https://jira.dhis2.org/browse/DHIS2-2255

This PR adds support for split view maps, which required quite a bit of refactoring of the Maps app. Support for timeline maps will come later in a separate PR after this is merged. 
- support for new "renderingStrategy" for layers. It is already supported in dhis2-core master (DHIS2-7186)
- New MapContainer component for the app which renders name, a MapView (below) and legend if in download mode. 
- New Plugin component which does the same as MapContainer for plugins. 
- New MapView component (shared by app and plugin) which decides on the rendering strategy. 
- Map component (for single maps) is refactored and now shared between app and plugin. 
- New SplitView component for layout.
- New MapItem component for maps that are part of a split view. 
- New PeriodName component for displaying periods on a MapItem
- New RederingStrategy component with radio buttons. 
- Changes to the thematic layer loader to support data loading for multiple periods separately. 
- Changes to map/ThematicLayers to only show values for a certain period. The period is also displayed in the popup. 
- New Legend component for plugin maps. Was previously in the GIS API, but legends should be independent of the engine in use. Was required to support a shared legend for split view. 
- New MapName component for plugins. Will make maps plugins more identical to chart and table plugins.
- If a split view layer is added, the user is restricted to add additional layers (message is shown in the add layer dialog)
- If other layers are added, "split view" is disabled and a message is shown in the layer dialog.
- Leaflet-relateded styles are either moved to GIS API, or made API agnostic. 
- Different map control configurations for single, split view and plugin maps (constants/mapControls.js)
- New wrapper around the map API which returns a map instance (makes it easier to switch the API in one place)

Other changes: 
- Upgrade Web API to v33. 
- Added support for passing analytical object id as a URL parameter (?id=sq23xXnqTF0) to plugin.html. It will load the map directly without the form. 

Puh!

Screenshot 2019-07-10 at 11 50 53
Screenshot 2019-07-10 at 11 49 32
Screenshot 2019-07-10 at 11 51 36
Screenshot 2019-07-10 at 11 49 58


turban:
I think the build is failing because of https://github.com/dhis2/gis-api/pull/58 - so we should merge that one first.

src/reducers/layerEdit.js
```diff
@@ -99,8 +99,10 @@ const layerEdit = (state = null, action) => {
         case types.LAYER_EDIT_PERIOD_TYPE_SET:
             return {
                 ...state,
-                periodType: action.periodType,
-                filters: removePeriodFromFilters(state.filters),
+                periodType: action.periodType.id,
+                filters: action.clearPeriod
```
amcgee: It looks like `clearPeriod` is always the default (`true`), what's the reason for having this condition here and adding the `clearPeriod` argument to the action?

amcgee:
Tested and found two bugs, sent videos in Slack:

1) If there's only one layer on a map, you can't edit the renderingMode of that layer even though there are no other layers present (it thinks there are)
2) This is a weird one - if you load an invalid orgunit (say the SL national boundary) and then edit that layer to use a valid orgunit (say "districts"), the maps still show as invalid.  Dragging on any map shows an error in the console.  Clicking on "zoom to extend" updates the map extent and shows the correct data, however panning with the mouse still doesn't work.  The same doesn't happen when loading the invalid orgunit in a non-split-view map.

Have a great weekend @turban !

src/components/map/Map.js
```diff
@@ -31,107 +22,83 @@ const layerType = {
...
 
-    componentDidMount() {
-        const { bounds, latitude, longitude, zoom } = this.props;
-        const map = this.context.map;
+    static childContextTypes = {
+        map: PropTypes.object.isRequired,
+    };
 
-        this.node.appendChild(map.getContainer()); // Append map container to DOM
+    getChildContext() {
```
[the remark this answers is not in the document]
turban: Yes it could, but I'm moving away from the MapProvider, as it don't work with split view which requires several map instances. The reason MapProvider is still is used is a simple check if a certain map layer is supported. My plan is to remove the MapProvider in a later PR.

src/components/layers/legend/LegendItemRange.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const LegendItemRange = ({ name = '', startValue, endValue, count }) => (
+    <td>
+        {isNaN(startValue) ? name : `${name} ${startValue} - ${endValue} `}
+        {count !== undefined ? `(${count})` : ''}
```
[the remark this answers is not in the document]
turban: Fixed.

src/components/map/SplitView.js
```diff
@@ -0,0 +1,93 @@
...
+                        />
+                        <PeriodName period={period.name} />
+                    </MapItem>
+                ))}
+            </div>
+        );
+    }
+
+    // Called from map child
+    setMapControls = map => {
```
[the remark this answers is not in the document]
turban: The main reason for this pattern is line 37 further up, where each control is moved from the map to the split view container element. Not super-clean, but it works :-) The split view container is created after the child components.

src/components/plugin/Legend.js
```diff
@@ -0,0 +1,122 @@
...
+    period: {
+        display: 'block',
+        fontWeight: 'normal',
+    },
+    alert: {
+        paddingBottom: 8,
+    },
+});
+
+class PluginLegend extends PureComponent {
```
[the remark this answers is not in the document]
turban: Fixed.

src/reducers/layerEdit.js
turban: It is set to false here: https://github.com/dhis2/maps-app/blob/feature/DHIS2-2255/src/components/periods/PeriodTypeSelect.js#L25

It was added to avoid clearing the Period dropdown when a favorite is loaded. It normally happens when the period type is changed.

turban: Added a comment in PeriodTypeSelect.js

turban:
Good catch @amcgee!

I've pushed fixes to the bugs you recorded. The first issue was fixed by passing in layerId to the RenderingStrategy component, so I could exclude the current layer in a better way.

The second issue was fixed (for now) by zooming the map to layer bounds every time the layer dialog is closed. We had it like this previously. It might be annoying for some user, so we might come up with a better solution (special case when no geometry are loaded previously) in a later PR.

src/components/map/Map.js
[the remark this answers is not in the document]
turban: I removed the MapProvider in this PR - it is obsolete.

src/components/layers/basemaps/BasemapList.js
```diff
@@ -11,13 +11,10 @@ const styles = {
...
 
-const BasemapList = (
-    { classes, selectedID, basemaps, selectBasemap },
-    { map }
-) => (
+const BasemapList = ({ classes, selectedID, basemaps, selectBasemap }) => (
     <div className={classes.container} data-test="basemaplist">
         {basemaps
-            .filter(basemap => map.hasLayerSupport(basemap.config.type))
+            // .filter(basemap => map.hasLayerSupport(basemap.config.type))
```
amcgee: Does this change functionality...?

src/components/map/ThematicLayer.js
```diff
@@ -71,9 +71,10 @@ class ThematicLayer extends Layer {
         map.addLayer(this.layer);
 
         // Only fit map to layer bounds on first add
-        if (!editCounter) {
-            this.fitBounds();
-        }
+        // Disabled due to issues when no geometries was found on first try
+        // if (!editCounter) {
```
amcgee: This should probably check the "current" bounds and call `this.fitBounds()` if this layer's bounds fall "outside" the current bounds, right?

Maybe we can do that in a separate change, but this is technically a regression as it stands right?

src/components/layers/basemaps/BasemapList.js
turban: No, this change was for Maps GL, and won't change the current functionality.

src/components/map/ThematicLayer.js
turban: I'll have a second look :-)

turban: Related PR:https://github.com/dhis2/gis-api/pull/59

I've added a new getZoom method that can be used to check if the map is already zoomed. If not, we will always try to fit to layer bounds when the layer is updated.

PR #155, opened 2019-07-20
feat: timeline support for thematic layers

turban:
JIRA-issue: https://jira.dhis2.org/browse/DHIS2-2255

This PR adds support for timeline/animated maps for thematic layers. 

It introduces a new Timeline component which is based D3.js components to make it adapt to various date ranges and map sizes.   

The timeline should show all periods returned from the backend, but the time axis only supports the English calendar. We should expand this when/if the Web API support different calendars: 
https://bl.ocks.org/mbostock/6f1cc065d4d172bcaf322e399aa8d62f
https://bl.ocks.org/mbostock/805115ebaa574e771db1875a6d828949

This PR also fixes some issues with map download introduced by #155. 

Screenshot 2019-07-22 at 09 42 58
Screenshot 2019-07-22 at 09 43 08
Screenshot 2019-07-22 at 09 42 21


turban:
I've done some serious testing, and added fixes to this PR :-)

src/components/map/Map.js
```diff
@@ -122,14 +122,10 @@ class Map extends Component {
 
     render() {
         const { basemap, layers, openContextMenu, classes } = this.props;
-        const overlays = [...layers.filter(layer => layer.isLoaded)].reverse();
```
amcgee: Does this not need to be reversed anymore?

amcgee:
Oh, and also will look into the failing tests

src/components/map/Map.js
turban: I did an extra check, and pushed a small fix. It is now rendering the layers correctly in both the app and the plugin. The layers are still reversed here: https://github.com/dhis2/maps-app/blob/feat/timeline/src/components/map/MapView.js#L40

turban:
Merging now, so I can do some extra testing on play tomorrow :-) Feel free to look at the cypress tests after your vacation :-)

amcgee:
The Cypress test was catching a legitimate error, an uncaught exception bubbles up and crashes the application

Screenshot 2019-07-26 14 06 55

This exists on the master branch now.  I will look into fixing it and do more testing while @turban is in the mountains!

PR #157, opened 2019-08-06
fix: re-export cypress fixtures

amcgee:
This fix updates requests and responses within Cypress fixtures which also fixes the Travis build failure on Master


amcgee:
I've set up Cypress Dashboard and made cypress test execution a bunch better.  However, there's currently a flakiness in the last Thematic Layer test because URLs are created non-deterministically.  The order of querystring parameters aren't consistent, so sometimes it misses the Cypress Fixture cache.  This is probably an issue with `d2` under the hood.

PR #162, opened 2019-08-14
fix: add babel plugin to transform destructuring

amcgee:
[DHIS2-7387](https://jira.dhis2.org/browse/DHIS2-7387)

This should fix a critical bug in the 2.32.1 release which renders maps-app unusable in IE11 (it renders as a blank screen, because there's a javascript syntax error in the bundle)

To fix this I specifically added a step to transpile any @dhis2-scoped packages in node_modules with babel in the webpack config


amcgee:
Travis CI appears to be failing because Cypress is broken on this branch.  I've done some work to improve it on master, but haven't backported.

If you add the URL to the play/dev CORS list you should be able to test this on netlify

PR #166, opened 2019-08-28
fix: bug fixes for Earth Engine layers

turban:
Fixes: 
https://jira.dhis2.org/browse/DHIS2-7344
https://jira.dhis2.org/browse/DHIS2-7345

- [x] Show year for temperature and precipitation
- [x] Show period for landcover
- [x] Update map when layer is edited
- [x] Toggle visibility and reorder individual EE layers


turban:
Merged in Maps 33.0.6, master and v33 branch.

PR #167, opened 2019-08-30
fix: try to restore Cypress functionality

amcgee:
This at least allows the Travis build to run the tests on v32, though there will still be race-condition flakiness when running the first Cypress test, so we might need to restart builds from time to time.


amcgee:
I updated the wait-on check to wait for a GET request instead of HEAD which **might** prevent the race condition!

PR #168, opened 2019-08-30
*WIP* Resize handling for map plugins

turban:


src/components/plugin/Plugin.js
```diff
@@ -63,6 +64,14 @@ class Plugin extends Component {
         );
     }
 
+    // Call this method when plugin container is resized
+    resize() {
+        // Will trigger a redraw of the MapView component
+        this.setState({
```
amcgee: To be safe this should use the [functional version](https://reactjs.org/docs/react-component.html#setstate) of `setState`, otherwise I think it could get a stale value.

```js
this.setState(state => ({ resizeCount: state.resizeCount + 1 }))
```

PR #178, opened 2019-09-18
fix: format coordinate values in event popups

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-6394

The Web API returns corrdinates as a string array with high decimal precision. The fix formats this value to use 6 decimals (meter precision). 

Needs to be backported. 

Skjermbilde 2019-09-18 kl  11 40 46

Skjermbilde 2019-09-18 kl  11 39 45

### DIFF
--- a/src/components/map/EventLayer.js
+++ b/src/components/map/EventLayer.js
@@ -157,7 +157,7 @@ class EventLayer extends Layer {
                 ' ' +
                 data.eventDate.substring(11, 16);
             const dataValues = data.dataValues;
-            let content = '<table><tbody>';
+            let content = '<div style="overflow-x:auto"><table><tbody>';
 
             // Output value if styled by data item, and item is not included in display elements
             if (styleDataItem && !this.displayElements[styleDataItem.id]) {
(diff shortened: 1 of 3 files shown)


src/components/map/EventLayer.js
```diff
@@ -169,13 +169,18 @@ class EventLayer extends Layer {
...
                     if (displayEl) {
-                        let value = dataValue.value;
-
-                        if (displayEl.optionSet) {
-                            value = displayEl.optionSet[value];
+                        const { valueType, optionSet, name } = displayEl;
+                        let { value } = dataValue;
+
+                        if (valueType === 'COORDINATE' && value) {
+                            value = JSON.parse(value)
```
amcgee: This will throw an error if the server doesn't return valid JSON, we should move it to another function and wrap it in a try/catch

src/components/map/EventLayer.js
```diff
@@ -169,13 +169,18 @@ class EventLayer extends Layer {
...
+                                .map(v => v.toFixed(6))
```
amcgee: Similarly, this assumes that the returned value is an array of integers, which could easily break and crash the app.  It should be at minimum try/catch-guarded

src/components/map/EventLayer.js
turban: Wrapped in a try/catch

src/components/map/EventLayer.js
turban: Wrapped in a try/catch

PR #180, opened 2019-09-19
fix: download multiple values for each feature

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-7556

Value for each period is included in the data download from split view or timeline maps. 

Skjermbilde 2019-09-19 kl  11 30 11


amcgee:
Posthumous approval!

PR #181, opened 2019-09-19
fix: loading indicator for new layers

turban:
Fixes an issue with https://github.com/dhis2/maps-app/pull/177 where the loading spinner was not showing for newly created layers. The layer don't exist in the redux store before it's loaded the first time, so it's not possible to set the loading state. As a fix, I've added a "newLayerIsLoading" property on the map object which is used in this case.

### DIFF
--- a/src/components/map/MapContainer.js
+++ b/src/components/map/MapContainer.js
@@ -36,6 +36,7 @@ const MapContainer = ({
     mapViews,
     bounds,
     showName,
+    newLayerIsLoading,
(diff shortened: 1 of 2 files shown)


turban:
TODO: newLayerIsLoading will not cover an edge case where another layer is created while the first is still loading. The only concequence would be that the spinner is removed before both layers are loaded, which will rarely happen.

A way forward would be to add layers to the redux store as they are created, and not wait for the initial data load.

PR #182, opened 2019-10-02
fix: refactor popup handling

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-7675

This PR refactors the popup handling in the Maps app, which is part of the larger Maps GL transition. Instead of generating HTML strings of popup content, we use react and createPortal from react-dom. 

A separate Popup component was created where the content is passed in as child elements. 

The event layer popup handling is complex, and an EventPopup component was created to separate it from the layer handling.

### DIFF
--- a/src/components/map/EventPopup.js
+++ b/src/components/map/EventPopup.js
@@ -144,15 +144,33 @@ const EventPopup = props => {
 
     // Load data elements every time programStage is changed
     useEffect(() => {
-        loadDataElements(programStage, eventCoordinateField).then(
-            setDataElements
-        );
-    }, [programStage]);
+        let aborted = false;
+
+        loadDataElements(programStage, eventCoordinateField).then(data => {
+            if (!aborted) {
+                setDataElements(data);
+            }
+        });
+
+        return () => {
+            aborted = true;
+        };
+    }, [programStage, eventCoordinateField, setDataElements]);
 
     // Load event data every time a new feature is clicked
     useEffect(() => {
-        loadEventData(feature).then(setEventData);
-    }, [feature]);
+        let aborted = false;
+
+        loadEventData(feature).then(data => {
+            if (!aborted) {
+                setEventData(data);
+            }
+        });
+
+        return () => {
+            aborted = true;
+        };
+    }, [feature, setEventData]);
 
     if (!coordinates || !feature || !eventData) {
         return null;
--- a/src/components/map/EventPopup.js
+++ b/src/components/map/EventPopup.js
@@ -1,17 +1,184 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-// import i18n from '@dhis2/d2-i18n';
+import i18n from '@dhis2/d2-i18n';
+import { getInstance as getD2 } from 'd2';
 import Popup from './Popup';
+import { apiFetch } from '../../util/api';
+import { getDisplayPropertyUrl, formatCoordinate } from '../../util/helpers';
 
-const EventPopup = props => {
-    const { coordinates } = props;
-    // console.log('EventPopup', props);
+class EventPopup extends PureComponent {
+    static propTypes = {
+        coordinates: PropTypes.array,
+        feature: PropTypes.object,
+        programStage: PropTypes.object,
+        eventCoordinateField: PropTypes.string,
+        styleDataItem: PropTypes.object,
+        onClose: PropTypes.func.isRequired,
+    };
 
-    return <Popup coordinates={coordinates}>###</Popup>;
-};
+    state = {
+        data: null,
+        displayElements: null,
+        eventCoordinateFieldName: null,
+    };
 
-EventPopup.propTypes = {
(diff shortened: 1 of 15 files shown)


turban:
Made a new commit to make the new Plugin component functional using react hooks.

src/components/map/Popup.js
```diff
@@ -0,0 +1,32 @@
+import { useEffect } from 'react';
+import { createPortal } from 'react-dom';
+import PropTypes from 'prop-types';
+import './Popup.css';
+
+const container = document.createElement('div');
```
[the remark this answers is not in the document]
turban: Fixed using a useMemo hook: 

const container = useMemo(() => document.createElement('div'), []);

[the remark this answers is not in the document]
turban: EventPopup is now functional and divided into smaller pieces/functions.

src/components/map/Popup.js
```diff
@@ -0,0 +1,31 @@
...
+const Popup = (props, context) => {
+    const { className = '', coordinates, onClose, children } = props;
+    const { map } = context;
+    const container = useMemo(() => document.createElement('div'), []);
+
+    useEffect(() => {
+        container.className = className;
+        map.openPopup(container, coordinates, onClose);
+        return () => map.closePopup();
+    });
```
amcgee: I think this needs an empty dependency array (`useEffect(..., [])`), otherwise this will be called on every render (rather than just on mount)

src/components/map/EventPopup.js
```diff
@@ -0,0 +1,216 @@
...
+    // Load data elements every time programStage is changed
+    useEffect(() => {
+        loadDataElements(programStage, eventCoordinateField).then(
+            setDataElements
+        );
+    }, [programStage]);
+
+    // Load event data every time a new feature is clicked
+    useEffect(() => {
+        loadEventData(feature).then(setEventData);
```
amcgee: To be pedantic - there is a race condition here.  If `loadEventData` is called and then the component unmounts before the promise resolves, the `then()` clause will still execute and try to set state on an unmounted component (an error/warning in React).  To avoid this we need to track an **aborted** state, something like this should work:

```js
useEffect(() => {
  let aborted = false;
  loadEventData(feature).then(data => {
    if (!aborted) {
      setEventData(data)
    }
  })
  return () => {
    aborted = true;
  }
}, [feature])
```

src/components/map/EventPopup.js
```diff
@@ -0,0 +1,216 @@
...
+    const [eventData, setEventData] = useState();
+    const [dataElements, setDataElements] = useState({});
+    const { displayElements = [], eventCoordinateFieldName } = dataElements;
+
+    // Load data elements every time programStage is changed
+    useEffect(() => {
+        loadDataElements(programStage, eventCoordinateField).then(
+            setDataElements
+        );
+    }, [programStage]);
```
amcgee: Same as `loadEventData`, this needs to be able to abort.

src/components/map/EventPopup.js
```diff
@@ -0,0 +1,216 @@
...
+
+    // Load event data every time a new feature is clicked
+    useEffect(() => {
+        loadEventData(feature).then(setEventData);
+    }, [feature]);
+
+    if (!coordinates || !feature || !eventData) {
```
amcgee: Should we move the `!feature || !eventData` check to be **inside** the Popup and show a loading spinner while the things are loading?  Otherwise there's a delay with nothing visible after the user clicks and before the popup displays

amcgee: Also, if coordinates is undefined nothing happens?  When will coordinates ever be undefined?

turban:
Thanks for good feedback @amcgee! Need to finalise this after India :-)

turban:
We need to improve the view of the popup when there a long data element names: 

Skjermbilde 2019-10-24 kl  21 25 10

A fix would be to allow word wrap, and alternate background for the rows.

src/components/map/Popup.js
turban: Solved by adding all dependencies to the array.

src/components/map/EventPopup.js
turban: Fixed.

src/components/map/EventPopup.js
turban: As this load should only happen once per layer (and not per popup), I've moved it back to the EventLayer component where I think it belongs.

src/components/map/EventPopup.js
turban: I've done some refactoring here and made feature and coordinates required props of the popup. My original plan was to create a shared popup component as the event layer was created, but itwas not very clean. Now a new popup is created every time the user clicks on a new event. 

I tested with a spinner in the popup, but I didn't like the effect. It results in a fast blinking popup, as the popup with a spinner is much smaller. We don't know the size of the popup before the data is loaded.

If needed, I think we should solve a popup loading indicator in a separate issue.

PR #256, opened 2019-10-24
chore(deps): upgrade d2-i18n(-generate) deps (master)

varl:


varl:
@turban close this if you don't want it.

I saw that you fixed the other branches, but master still has d2-i81n@1.0.4 and d2-i18n-generate@1.0.19, this fixes that.

PR #258, opened 2019-10-24
Show alert if no boundaries was returned

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-7773

The problem was that we didn't return any layer config if the array was empty. The code works fine empty arrays, so the fix was to remove an early return, and then include an alert if no boundary features was retuned (= empty array). 

Skjermbilde 2019-10-24 kl  20 22 31

### DIFF
--- a/src/components/map/EventLayer.js
+++ b/src/components/map/EventLayer.js
@@ -157,7 +157,8 @@ class EventLayer extends Layer {
                 ' ' +
                 data.eventDate.substring(11, 16);
             const dataValues = data.dataValues;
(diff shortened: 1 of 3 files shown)


turban:
Added check for data elements in infrastructural dialog

turban:
Added a max height to event popups. I think this fix is sufficient for released versions. Will improve the look in this event popup PR for future versions: https://github.com/dhis2/maps-app/pull/182

amcgee:
> Added a max height to event popups. I think this fix is sufficient for released versions. Will improve the look for future Maps app versions in this PR: #182

Agreed, sounds good!

PR #263, opened 2019-11-14
fix: clear dimension items array when filter is changed

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-7866

When the user changed filter dimension, we did not clear the selected dimension items, which resulted in a crash.

### DIFF
--- a/src/components/dimensions/DimensionFilterRow.js
+++ b/src/components/dimensions/DimensionFilterRow.js
@@ -46,7 +46,9 @@ class DimensionFilterRow extends Component {
             <div className={classes.container}>
                 <DimensionSelect
                     dimension={dimension}
-                    onChange={dimension => this.onChange(dimension.id)}
+                    onChange={selectedDimension =>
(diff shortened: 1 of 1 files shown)


src/components/dimensions/DimensionFilterRow.js
```diff
@@ -46,7 +46,7 @@ class DimensionFilterRow extends Component {
             <div className={classes.container}>
                 <DimensionSelect
                     dimension={dimension}
-                    onChange={dimension => this.onChange(dimension.id, items)}
+                    onChange={dimension => this.onChange(dimension.id)}
```
amcgee: Can we change the name of this variable to `dimensionSelectEntry` or something?  It's quite confusing looking at this function, since `dimension` in the outer scope is the dimension `id` but `dimension` in the inner scope is an object, and both are passed at different times to `onChange`

turban: Agreed. Changed "selectedDimension".

PR #404, opened 2019-12-10
feat: support for bing maps and donut clusters

turban:
This PR includes various changes to make DHIS2 Maps compliant with Maps GL:
- Support for base maps from Bing. 
- Only show base maps supported by the map api in use. 
- Add support for "donut clusters". 
- New offset argument for popups (used when clusters are "spiderified").
- Use numeric values for GeoJSON feature ids (limited string support in Mapbox GL JS).
- Align server clusters with the Mapbox GL cluster format.

The PR includes a Bing Maps key which will later be stored in the instance db. The key will be removed from this repo and deleted from Bing when backend support is in place.


src/components/map/EventLayer.js
```diff
@@ -119,11 +131,13 @@ class EventLayer extends Layer {
         ) : null;
     }
 
-    onEventClick({ feature, coordinates }) {
-        this.setState({ popup: { feature, coordinates } });
+    onEventClick({ feature, coordinates, offset }) {
+        this.setState({ popup: { feature, coordinates, offset } });
```
amcgee: Question: where does this offset come from?

src/components/map/EventLayer.js
```diff
@@ -136,17 +150,23 @@ class EventLayer extends Layer {
...
+                        cluster_id: clusterId,
+                        point_count: count,
+                        point_count_abbreviated: formatCount(count),
                         bounds: [
-                            [extent[1], extent[0]],
-                            [extent[3], extent[2]],
+                            [extent[0], extent[1]],
+                            [extent[2], extent[3]],
                         ],
+                        id: row[header.points],
```
amcgee: Doesn't this need to be `[EVENT_ID_FIELD]: row[header.points]`?  Or am I missing something?

src/components/map/EventLayer.js
turban: It comes from Maps GL: https://github.com/dhis2/maps-gl/blob/feat/event-layer/src/layers/Spider.js#L82

It is the offset from the spiderified marker to the clicked feature which don't have a coordinate.

src/components/map/EventLayer.js
turban: I don't think so as this is server cluster that don't make sense to download. I think you added [EVENT_ID_FIELD] for download purposes.

src/components/map/EventLayer.js
amcgee: OK, that's fine I just wasn't sure where it came from

src/components/map/EventLayer.js
amcgee: Maybe true, but in that case I think the below `feature.properties.id || feature.properties[EVENT_ID_FIELD];` might break data download, since we're now always setting `.id`?  I don't remember how this logic worked off the top of my head, can look into it if you want.

src/components/map/EventLayer.js
turban: Smaller API footprint is better, so I've now removed the need for the offset parameter in Maps GL. The solution was to add the offset to the [lng, lat] position directly and return it with the click event.

src/components/map/EventLayer.js
turban: id is only set for server clusters, which we don't download - when the user download a server cluster layer we make new api requests and fetches all the events. I've tested that download still works like before, but feel free to do some additional testing :-)

PR #405, opened 2019-12-11
fix: update headerbar to latest and greatest

varl:
I had to run prettify on the source to get a commit in, and it dragged along a lot of changes. All the functional changes are in `package.json` and `App.js`.


varl:
Ah, sorry. I agree and will fix.

PR #412, opened 2020-01-23
feat: mapping interface locale

turban:
This PR adds support for locale strings for map controls. 

It builds on the locale support in Mapbox GL JS:
https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/default_locale.js
https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L1237-L1244


turban:
@amcgee pushed a fix that includes the strings in i18n

turban:
@amcgee translation strings are now generated.

PR #434, opened 2020-01-30
fix: authentication header for map plugin

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-8215

Allows the user to authenticate with username and password with the map plugin.


amcgee:
I'd rather we leave this broken, actually, as it's a security hole... is that possible?

PR #446, opened 2020-02-11
fix: upgrade to v34 of backend api

turban:
This PR upgrades to v32 of the web api + minor class name change.


amcgee:
@turban are there v34 API features that don't exist in v33 and are needed by the maps app?

turban:
@amcgee we're not creating a breaking API change that I'm aware of. The new bing maps api key will probably be available for v33 as well.

PR #474, opened 2020-03-02
fix: Bing Maps download and more compact select lists

turban:
This PR fixes:
- Bing Maps key is loaded before a saved map to make sure if have the Bing Maps key available.
- Select lists are more compact. 
- Bing Maps logo is not included in map download (it's blocked by CORS policy).
- Allow maps to be downloaded if no title in present. 
- Set max radius for legend items

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -66,12 +66,14 @@ getManifest('manifest.webapp')
     })
     .then(configI18n)
     .then(init)
+    .then(d2 =>
+        d2.system.settings.get('keyBingMapsApiKey').then(key => {
+            store.dispatch(setBingMapsApiKey(key));
+            return d2;
+        })
+    )
     .then(
-        async d2 => {
-            await d2.system.settings
-                .get('keyBingMapsApiKey')
-                .then(key => store.dispatch(setBingMapsApiKey(key)));
-
+        d2 => {
             const mapId = getUrlParameter('id');
             if (mapId) {
                 store.dispatch(loadFavorite(mapId));
--- a/src/app.js
+++ b/src/app.js
@@ -67,8 +67,8 @@ getManifest('manifest.webapp')
     .then(configI18n)
     .then(init)
     .then(
-        d2 => {
-            d2.system.settings
+        async d2 => {
+            await d2.system.settings
                 .get('keyBingMapsApiKey')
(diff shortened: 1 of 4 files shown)


src/app.js
```diff
@@ -67,8 +67,8 @@ getManifest('manifest.webapp')
     .then(configI18n)
     .then(init)
     .then(
-        d2 => {
-            d2.system.settings
+        async d2 => {
```
amcgee: This is technically OK, but a bit weird to have a Promise `.then` within an `async/await` within another `.then`... why not either rewrite the whole function as in `async/await` style (cleaner) or keep the existing `.then` style with chained promises (smaller change)?

For the second option it would look something like this, since returning a promise in a `.then` callback will await that promise and await the result before calling the next `.then` callback: 

```js
.then(d2 =>
    d2.system.settings
        .get('keyBingMapsApiKey')
        .then(key => {
            store.dispatch(setBingMapsApiKey(key)))
            return d2 /* this could be avoided by using `getInstance()` below, but this is the existing behavior */
        })
).then(d2 => {
  const mapId = getUrlParameter('id')
  // ... etc ...
})
```

If you prefer the `async/await` addition as-is that's OK too, but in that case it should use the `await` rather than another `.then` inside the `async` function:

```js
    const key = await d2.system.settings.get('keyBingMapsApiKey')
    store.dispatch(setBingMapsApiKey(key))
```

src/components/layers/legend/LegendItem.js
```diff
@@ -23,9 +25,11 @@ const LegendItem = ({
     };
 
     if (radius) {
-        symbol.width = radius * 2 + 'px';
-        symbol.height = radius * 2 + 'px';
-        symbol.borderRadius = radius ? '50%' : '0';
+        const r = (radius < maxRadius ? radius : maxRadius) * 2;
```
amcgee: Should this just be `Math.min(radius, maxRadius) * 2`?

src/app.js
turban: Thanks for review @amcgee! I agree that it's more readable - I've pushed a change that aligns with your first suggestion.

src/components/layers/legend/LegendItem.js
turban: True, fixed!

PR #496, opened 2020-03-17
fix: Handle error if org unit selection is invalid for thematic layers (DHIS2-8479)

turban:
Partly fixes:  https://jira.dhis2.org/browse/DHIS2-8479

Related to the same issue with invalid org unit selections for facility layers: https://github.com/dhis2/maps-app/pull/495

The error originates in an invalid org unit selection that was not handled properly. 

Screenshot 2020-03-17 at 14 41 27

This PR includes a fix that check if period(s) are available (which is not the case if org unit selection is invalid). We allow the limited layer config to pass through so the user can edit the layer and fix the org unit selection. 

The error message received could be more descriptive, but I think it will work until we get new org uni selection in place: https://jira.dhis2.org/browse/DHIS2-8481

Screenshot 2020-03-17 at 14 42 24

### DIFF
--- a/src/components/map/Layer.js
+++ b/src/components/map/Layer.js
@@ -45,7 +45,7 @@ class Layer extends PureComponent {
             editCounter,
             dataFilters,
         } = this.props;
-        const { period } = this.state || {};
+        const { period } = this.state;
         const { period: prevPeriod } = prevState || {};
         const isEdited = editCounter !== prevProps.editCounter;
 
--- a/src/components/map/Layer.js
+++ b/src/components/map/Layer.js
@@ -27,6 +27,8 @@ class Layer extends PureComponent {
         isVisible: true,
     };
 
+    state = {};
+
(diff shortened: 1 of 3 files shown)


src/components/map/Layer.js
```diff
@@ -27,6 +27,8 @@ class Layer extends PureComponent {
         isVisible: true,
     };
 
+    state = {};
```
martinkrulltott: Makes this null value check redundant? https://github.com/dhis2/maps-app/blob/616f5f27deb1658a62a27e012a701a024ccdc474/src/components/map/Layer.js#L48

martinkrulltott:
Nice description, images and links btw 👍 👍

src/components/map/Layer.js
turban: Yes, that check is no longer needed. Now remoived :-)

PR #498, opened 2020-03-18
fix: set the default classification/legend for selected data item (DHIS2-8485)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-8485

This PR fixes an issue with thematic layer not getting default classification/legend stored if the styled tab is not visited by the user before "Add layer" button is clicked. 

The solution is to set the default classification/legend whenever a data item is first selected or changed.

Before this fix making this selection would produce a map with a predefined legend: 

Screenshot 2020-03-18 at 12 51 08
 
But the predefined legend was not saved with the map, as the selection was never made in the UI without first visiting the Style tab. This PR makes sure this is happening.

### DIFF
--- a/src/components/edit/thematic/ThematicDialog.js
+++ b/src/components/edit/thematic/ThematicDialog.js
@@ -117,7 +117,6 @@ export class ThematicDialog extends Component {
         labelFontStyle: PropTypes.string,
         labelFontWeight: PropTypes.string,
         legendSet: PropTypes.object,
-        method: PropTypes.number,
         indicatorGroup: PropTypes.object,
         dataElementGroup: PropTypes.object,
         program: PropTypes.object,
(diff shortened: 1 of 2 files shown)


src/components/edit/thematic/ThematicDialog.js
```diff
@@ -213,6 +223,21 @@ export class ThematicDialog extends Component {
                 .forEach(ou => loadOrgUnitPath(ou.id));
         }
 
+        // Set the default classification/legend for selected data item without visiting the style tab
+        if (columns !== prev.columns) {
+            const dataItem = getDataItemFromColumns(columns);
+
+            if (dataItem) {
```
martinkrulltott: in which case will this be false?

turban: It will be false when item type, or item group is changed.

PR #582, opened 2020-04-14
fix: clean ou dimension before app switching

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-8585

This PR fixes an issue where we try to serialise OU dimensions represented as models resulting in a circular structure error.  This fix is to use the same approach as we do on save, to clean the OU dimensions by only allowing the needed properties to pass through. 

Solves this error: 
image-2020-04-02-13-19-05-669

### DIFF
--- a/src/util/favorites.js
+++ b/src/util/favorites.js
@@ -167,7 +167,7 @@ const models2objects = config => {
     return config;
 };
 
-const cleanDimension = dim => ({
+export const cleanDimension = dim => ({
     ...dim,
     items: dim.items.map(item => pick(validModelProperties, item)),
 });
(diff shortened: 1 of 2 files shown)


src/util/favorites.js
```diff
@@ -167,7 +167,7 @@ const models2objects = config => {
     return config;
 };
 
-const cleanDimension = dim => ({
+export const cleanDimension = dim => ({
```
martinkrulltott, line 170: is this really located in the right place, now that it's being used by both `favourites` and `analyticalObject.js`?

turban: Yes, it could be moved to the util/analytics.js, but it requires a bigger rewrite I would like to save for later (and possibly reuse code you've made for DV).

PR #639, opened 2020-04-30
fix: make all strings translatable 2.32 (DHIS2-8658)

turban:
Backport to 2.33.4 of: https://jira.dhis2.org/browse/DHIS2-8658

Original PR: #590

A wrong import was found in the original PR, which is fixed in this PR and in #630


turban:
Well spotted @martinkrulltott! Yes the logic around the PeriodTypeSelect was changed in 2.33 when we added support for timeline and split view maps.

PR #657, opened 2020-05-05
fix: layer name translations and d2-ui upgrades

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-8658?focusedCommentId=33085&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-33085

Added some missing translation handing for layer names. The default layers defined in reducers/layers.js is evaluated before the i18n knows the language in use, so I had to add a `i18n.t(type)` in components/layers/layers/Layer.js. Not very clean, but this should work fine until we switch to the app plattform with better i18n handling in 2.35. 

Upgraded d2-ui-file-menu and d2-ui-interpretations to version 7.0.0 for translation fixes.


src/components/layers/layers/Layer.js
```diff
@@ -45,8 +46,8 @@ const styles = theme => ({
 
 const Layer = ({ classes, layer, onClick }) => {
     const { img, type, name } = layer;
+    const label = name || i18n.t(type);
```
jenniferarnesen, line 49: Does i18n.t(type) actually translate correctly? I was pretty sure you had to insert an actual string.

PR #697, opened 2020-05-12
fix: add missing translations (DHIS2-8658)

turban:
This PR adds i18n support for strings that couldn't be translated in the maps app. 

`i18n.t(name)` will not be discovered by the translation system, but all names are translated in basemaps.js, but this reducer is evaluated before the language setting is available. Will fix better when we switch to the app plattform. 

`subtitle: 'Basemap'` is removed as this is the default value defined here: https://github.com/dhis2/maps-app/blob/master/src/components/layers/basemaps/BasemapCard.js#L71

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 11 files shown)


turban:
Good idea @martinkrulltott! I'll include this check and screenshots in future i18n PRs. It made me find a missing translation, now fixed! 

Screenshot 2020-05-13 at 10 47 34

Screenshot 2020-05-13 at 10 48 36

Screenshot 2020-05-13 at 10 47 54

Screenshot 2020-05-13 at 10 46 57

PR #783, opened 2020-05-26
fix: upgrade headerbar to support translations - 2.32 backport

turban:
2.32 backport of #781 and parts of #405 to be able to upgrade headerbar to support translations.


turban:
@amcgee it's a bit strange as the polyfill is present in 2.34, but not 2.33 which still passed the checks. I don't know why.

PR #985, opened 2020-08-21
fix: require legend set selection (DHIS2-9314)

turban:
If the user selects "Predefined color legend" it is also required to select a legend set. 

Fixes: https://jira.dhis2.org/browse/DHIS2-9314

Screenshot 2020-08-21 at 22 26 00


edoardo (approved):
LGTM

PR #993, opened 2020-08-25
Show TEI relationships outside a program (DHIS2-8715)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-8715

This PR fixes two issues: 

**1. Allow user to select between "last updated" and "program" periods**
Currently we use "last updated" periods when no program is selected and program/enrollment periods when we have a program. With this PR we always use "last updated" dates by default, but the user can switch to program dates if applicable. 

When no program is selected: 
Screenshot 2020-08-25 at 15 25 03

When a program is selected: 
Screenshot 2020-08-25 at 15 25 17

**2. Show TEI relationships outside program**
Currently, we don't show TEI relationships to instances outside a program, if a program is selected. In this PR, a checkbox is added to allow this: 

Screenshot 2020-08-25 at 15 29 19

If checked the relationship will show on the map: 

Screenshot 2020-08-25 at 15 31 46

Before this PR only the top red circle will show, as the related instance is not in the same program. 

TODO: 
- Allow TEI period type and and TEI relationships setting to be saved


src/util/teiRelationshipsParser.js
```diff
@@ -52,6 +57,15 @@ const isIndexInstance = (instance, type) => {
     return hasChildren;
 };
 
+const getTargetInstances = soureInstances =>
```
amcgee, line 60: Typo - should be `sourceInstances`

amcgee (approved):
One minor comment but code LGTM.  I haven't tested.

I think the messaging in the UI could be a little cleaner, but will leave that to @cooper-joe

src/util/teiRelationshipsParser.js
turban: Fixed!

cooper-joe:
> I think the messaging in the UI could be a little cleaner, but will leave that to @cooper-joe

I agree, I'm going to try to come up with something clearer - but it's quite an abstract concept so I'm not sure we'll be able to. I think this can be merged for functionalities sake though.

cooper-joe:
I took a quick look at the ui text now, I suggest this adjustment:
period-toggle


I think the text is clearer here, what do you think @turban. I'm not 100% sure I'm communicating the 'Program/Enrollment' date. I'm not sure it is always referred to as 'Enrollment' date 🤔

turban:
I think it's clearer @cooper-joe, and I included your suggestion before the merge.

PR #1083, opened 2020-09-23
fix: unique class name prefix for each map plugin item (DHIS2-9593)

turban:
This PR fixes an issue with colling class names when having multiple maps on the same dashboard. 

Fixes: https://jira.dhis2.org/browse/DHIS2-9593

The solution is to include an unique map id in the class prefix. The default is to use the id from the analytical object, if undefined, a new id will be generated. 

Including rendering method in the class prefix is obsolete when using unique ids, so it was removed. 

The plan is to only merge this fix to master and test it thoroughly on play dev before backporting. 

This is related to an issue with the styling solution in Material UI described here (mutiple react contexts): 
https://github.com/mui-org/material-ui/issues/8872#issuecomment-340043309

The plan is to stop using Material UI in future versions of DHIS2 Maps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "maps-app",
-    "version": "35.0.10",
+    "version": "35.0.11",
     "description": "DHIS2 Maps",
     "main": "src/app.js",
     "repository": {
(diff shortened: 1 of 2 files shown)


janhenrikoverland (approved):
LG

PR #1095, opened 2020-09-25
fix: show TEIs regardless of relationship

turban:
If the user checks "Display Tracked Entity relationships" only TEIs having relationships will show on the map. This is not obvious for the user, and this fix will always show the same TEIs regardless if they take part in a relationship or not. 

The solution is to draw all TEIs (as we do when no relationships are shown). If a TEI take part as a secondary instance in a relationship (B) we will not draw it twice. Primary TEIs (A) and TEIs not present in a relationship will have the same style.   

Fixes: https://jira.dhis2.org/browse/DHIS2-9622

### DIFF
--- a/src/loaders/trackedEntityLoader.js
+++ b/src/loaders/trackedEntityLoader.js
@@ -134,8 +134,7 @@ const trackedEntityLoader = async config => {
         );
     }
 
-    const data = toGeoJson(instances);
-    let relationships, secondaryData;
+    let data, relationships, secondaryData;
 
     if (relationshipTypeID) {
         const relationshipType = await apiFetch(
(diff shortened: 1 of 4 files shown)


src/loaders/trackedEntityLoader.js
```diff
@@ -170,7 +170,13 @@ const trackedEntityLoader = async config => {
...
         );
 
-        data = toGeoJson(dataWithRels.primary);
+        // Show all TEIs expect those that are shown as secondary instances
+        data = toGeoJson(
+            instances.filter(
+                tei => !dataWithRels.secondary.find(s => s.id === tei.id)
+            )
+        );
+
```
amcgee, line 179: I think this should work, but I'm a little worried about performance - this has O(n^2) complexity which isn't great.  Should we remove the `primary` array calculation or update it to include instances without relationships?  In any case, we could probably improve performance later if it's important to get this in.

amcgee (approved):

turban:
Closing this pull request as the code is not stabel for 2.35.0 release.

PR #1102, opened 2020-09-29
Remove functionality to show TEI relationships outside program

turban:
Partly reverts changes from #993 (2. Show TEI relationships outside program)

This functionality was made as a change just before 2.35 soft freeze and not well tested, partly because of limited demo data. 
The code has bugs: https://jira.dhis2.org/browse/DHIS2-9634

It is safer to remove the support for TEI relationships outside program than to fix it just before a major release, as the implementation is quite complex. 

Most of the functionality is kept in code, so it can be enabled again at a later stage.


amcgee (approved):
Code change LGTM 👍

PR #1109, opened 2020-10-05
feat: replace Material UI with @dhis2/ui (DHIS2-9699)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-9699


package.json
```diff
@@ -93,6 +92,11 @@
         "url-polyfill": "^1.1.10",
         "whatwg-fetch": "^3.4.0"
     },
+    "peerDependencies": {
+        "react": "^16.13.1",
+        "react-dom": "^16.13.1",
+        "styled-jsx": "^3.3.0"
+    },
```
amcgee, line 99: I don't think these should be peerDependencies, since this is the top-level app.  There may be some straggling documentation somewhere which suggests using peerDeps, but regular dependencies work fine!

amcgee (commented):
Sorry for the premature review, just saw this as I was looking around!  Glad the prop-types build issue was resolved :-)

package.json
turban: Read it here: https://ui.dhis2.nu/#/getting-started
Will fix!

PR #1139, opened 2020-10-12
chore: switch to material ui to @dhis2/ui for download map dialog (DHIS2-9699)

turban:

This PR includes the setup for @dhis2/ui and CSS Modules, and replaces Material UI components in the download map dialog. 

The default test environment is used for Jest (not node), as @dhis2/ui requires access to more DOM features. 

With this PR the download map dialog looks like this: 

Screenshot 2020-10-12 at 15 12 37

Before this PR: 

Screenshot 2020-10-12 at 13 24 57


cooper-joe (approved):
LGTM!

src/components/download/DownloadDialog.js
```diff
@@ -78,29 +64,34 @@ export class DownloadDialog extends Component {
...
-                </DialogTitle>
-                <DialogContent className={classes.content}>
+            <Modal
+                open={showDialog}
+                position="middle"
+                small
+                onClose={this.onClose}
+            >
+                <ModalTitle>{i18n.t('Download map')}</ModalTitle>
+                <ModalContent className={styles.modalContent}>
```
edoardo, line 74: I'm unsure about this way of customising styles.
See: https://ui.dhis2.nu/#/advanced-usage

src/components/download/LegendPosition.js
```diff
@@ -10,64 +10,18 @@ export const legendPositions = [
...
-        <div className={classes.label}>{i18n.t('Legend position')}</div>
+export const LegendPosition = ({ position, onChange }) => (
+    <div className={styles.root}>
+        <div className={styles.label}>{i18n.t('Legend position')}</div>
         {legendPositions.map(pos => (
             <div
                 key={pos}
-                className={`${classes.position} ${
-                    pos === position ? classes.selected : ''
+                className={`${styles.position} ${
```
edoardo, line 19: Here you can use `cx` for composing the class names, it's a bit cleaner and we used it around.

src/components/download/DownloadDialog.js
turban: Fixed by adding class names to a div rather than the UI element.

src/components/download/LegendPosition.js
turban: Switched to cx - thanks for the advice!

src/components/download/DownloadDialog.js
```diff
@@ -78,57 +64,64 @@ export class DownloadDialog extends Component {
...
-                    </Button>
-                    {isSupported && (
-                        <Button
-                            variant="contained"
-                            color="primary"
-                            onClick={this.onDownload}
-                        >
-                            {i18n.t('Download')}
+            <Modal
+                open={showDialog}
```
martinkrulltott, line 68: I don't think `Modal` has an `open` prop, usually we use conditional rendering instead

src/components/download/DownloadDialog.js
```diff
@@ -78,57 +64,64 @@ export class DownloadDialog extends Component {
...
+                position="middle"
+                small
+                onClose={this.onClose}
+            >
+                <ModalTitle>{i18n.t('Download map')}</ModalTitle>
+                <ModalContent>
+                    <div className={styles.modalContent}>
+                        {isSupported ? (
+                            <Fragment>
```
martinkrulltott, line 77: Consider using the `<>` [shorthand](https://reactjs.org/docs/fragments.html#short-syntax) instead

src/components/download/LegendPosition.js
```diff
@@ -10,64 +11,18 @@ export const legendPositions = [
...
-        right: 8,
-    },
-});
-
-export const LegendPosition = ({ position, onChange, classes }) => (
-    <div className={classes.root}>
-        <div className={classes.label}>{i18n.t('Legend position')}</div>
+export const LegendPosition = ({ position, onChange }) => (
+    <div className={styles.root}>
+        <div className={styles.label}>{i18n.t('Legend position')}</div>
```
martinkrulltott, line 16: From a HTML semantics perspective I'd consider using something else than a `div` to wrap the label

src/components/download/__tests__/LegendPosition.spec.js
```diff
@@ -3,17 +3,10 @@ import { shallow } from 'enzyme';
...
-        topright: 'topright-class',
-        bottomright: 'bottomright-class',
-        selected: 'selected-class',
-    };
-
     const renderComponent = props =>
         shallow(
             <LegendPosition
-                classes={classes}
+                // classes={classes}
```
martinkrulltott, line 9: Should be deleted 👍

martinkrulltott (approved):
Looks great. Just a few tiny comments, nothing critical 👍 🥇  Happy to approve as is.

src/components/download/DownloadDialog.js
turban: I'm getting errors in my testes if I use the shorthand (seems like I need to upgrade til Babel 7, which will happen when we move to the app platform).

edoardo (approved):

src/components/download/LegendPosition.js
turban: Switched to <label> element https://html.spec.whatwg.org/multipage/forms.html#the-label-element

src/components/download/DownloadDialog.js
turban: Fixed.

src/components/download/__tests__/LegendPosition.spec.js
turban: Fixed.

PR #1146, opened 2020-10-13
chore: switch to material ui to @dhis2/ui for org unit dialog (DHIS2-9699)

turban:

This PR replaces Material UI components and styles in the org unit dialog (right click an org unit and select "Show information"). The table from @dhis2/ui is used to show infrastructural data. 

After this PR: 

Screenshot 2020-10-13 at 22 00 17

Before this PR:

Screenshot 2020-10-13 at 22 00 26

### DIFF
--- a/src/components/orgunits/OrgUnitDialog.js
+++ b/src/components/orgunits/OrgUnitDialog.js
@@ -7,16 +7,10 @@ import {
     ModalTitle,
     ModalContent,
     ModalActions,
-    Table,
-    TableHead,
-    TableRowHead,
-    TableCellHead,
-    TableBody,
-    TableRow,
-    TableCell,
     Button,
 } from '@dhis2/ui';
 import PeriodSelect from '../periods/PeriodSelect';
+import OrgUnitDialogTable from './OrgUnitDialogTable';
 import { closeOrgUnit } from '../../actions/orgUnits';
 import { loadConfigurations, loadData } from '../../util/infrastructural';
 import styles from './styles/OrgUnitDialog.module.css';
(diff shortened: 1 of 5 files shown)


src/components/orgunits/OrgUnitDialog.js
```diff
@@ -130,75 +85,78 @@ export class OrgUnitDialog extends Component {
...
+                                                <TableCell
+                                                    dense
+                                                    className={styles.right}
+                                                >
+                                                    {value}
+                                                </TableCell>
+                                            </TableRow>
+                                        ))}
+                                    </TableBody>
+                                </Table>
```
amcgee, line 138: Random suggestion - can we move the table out into its own component?

cooper-joe:
Would it be possible to use a `ui` `SingleSelectField` ([example](https://ui.dhis2.nu/demo/?path=/story/singleselectfield--default)) for the period selection? I'm not sure if it would work out of the box with the left/right arrows for moving between periods.

turban:
> Would it be possible to use a `ui` `SingleSelectField` ([example](https://ui.dhis2.nu/demo/?path=/story/singleselectfield--default)) for the period selection? Although I'm not sure if it would work out of the box with the left/right arrows for moving between periods.

Yes, it will be fixed in a separate PR.

src/components/orgunits/OrgUnitDialog.js
turban: Done!

cooper-joe:
> > Would it be possible to use a `ui` `SingleSelectField` ([example](https://ui.dhis2.nu/demo/?path=/story/singleselectfield--default)) for the period selection? Although I'm not sure if it would work out of the box with the left/right arrows for moving between periods.
> 
> Yes, it will be fixed in a separate PR.

Ok 👍 LGTM

cooper-joe (approved):

PR #1147, opened 2020-10-14
chore: switch to @dhis2/ui tabs (DHIS2-9699)

turban:

This PR replaces Material UI tab components with equivalents from @dhis2/ui. Some wrapper components was created to have a single onChange handler which simplifies the code in the layer dialogs. Please note that only the tabs are updated, there are still several Material UI components in the layer dialogs (will be fixed in separate PRs). 

After this PR:

Screenshot 2020-10-14 at 19 00 57
Screenshot 2020-10-14 at 19 01 11
Screenshot 2020-10-14 at 19 01 24
Screenshot 2020-10-14 at 19 01 38
Screenshot 2020-10-14 at 19 01 55
Screenshot 2020-10-14 at 19 02 08

Before this PR: 

Screenshot 2020-10-14 at 19 03 24
Screenshot 2020-10-14 at 19 03 35
Screenshot 2020-10-14 at 19 03 48
Screenshot 2020-10-14 at 19 04 01
Screenshot 2020-10-14 at 19 04 13
Screenshot 2020-10-14 at 19 04 22

### DIFF
--- a/src/components/core/Tab.js
+++ b/src/components/core/Tab.js
@@ -1,15 +1,29 @@
-import React from 'react';
-import { withStyles } from '@material-ui/core/styles';
-import { Tab as MuiTab } from '@material-ui/core';
+import React, { useContext, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { Tab as UiTab } from '@dhis2/ui';
+import { TabContext } from './Tabs';
 
-const styles = {
-    root: {
-        minWidth: 0,
-        fontSize: 14,
-    },
+const Tab = ({ value, dataTest, children }) => {
+    const { tab, onChange } = useContext(TabContext);
+
+    // onChange is from the parent component
+    const onClick = useCallback(() => {
+        if (value !== tab) {
+            onChange(value);
+        }
+    }, [value, tab, onChange]);
+
+    return (
+        <UiTab selected={value === tab} onClick={onClick} dataTest={dataTest}>
+            {children}
+        </UiTab>
+    );
 };
 
-// Styled wrapper around MUI Tab
-const Tab = props => <MuiTab {...props} />;
+Tab.propTypes = {
+    value: PropTypes.string.isRequired,
+    dataTest: PropTypes.string,
+    children: PropTypes.node.isRequired,
+};
 
-export default withStyles(styles)(Tab);
+export default Tab;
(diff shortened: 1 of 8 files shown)


src/components/core/Tab.js
```diff
@@ -1,15 +1,29 @@
...
+import React, { useContext, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { Tab as UiTab } from '@dhis2/ui';
+import { TabContext } from './Tabs';
+
+const Tab = ({ value, dataTest, children }) => {
+    const { tab, onChange } = useContext(TabContext);
+
+    // onChange is from the parent component
+    const onClick = useCallback(() => {
```
edoardo, line 10: I'm not very familiar with `useCallback` yet.
Is this really required here?

src/components/edit/BoundaryDialog.js
```diff
@@ -118,11 +118,8 @@ class BoundaryDialog extends Component {
         return (
             <div data-test="boundarydialog">
                 <Tabs value={tab} onChange={tab => this.setState({ tab })}>
```
edoardo, line 120: I wonder if this can be simplified by using directly `TabBar` and `Tab` from `ui` in these places.
The custom components `Tabs` and `Tab` don't seem to do anything special that won't work with the `ui` counterparts?
Maybe wrapping was necessary before because of MUI? Or was it a choice to wrap everything to not bind Maps app to any particular UI library and allow for easy replacement of the underlying UI lib?

cooper-joe:
The switch to `ui` looks good to me. 

My only comment is whether we could remove the single 'style' tab on Earth Engine layer? Tabs should only really be used if there is two or more of them. Perhaps out of scope of this PR, but I would suggest that change if there is never a second/third tab available for earth engine layers.

cooper-joe (approved):

edoardo (approved):
Overall LGTM, but I'm wondering if we want to simplify the use of wrappers for components?

src/components/edit/BoundaryDialog.js
turban: TabBar don't have an onChange handler, so each tab needs a selected state and an onClick handler, so I felt this was simplifying the code on the layer dialogs.

turban:
> My only comment is whether we could remove the single 'style' tab on Earth Engine layer? Tabs should only really be used if there is two or more of them. Perhaps out of scope of this PR, but I would suggest that change if there is never a second/third tab available for earth engine layers.

Fixed!

src/components/core/Tab.js
turban: useCallback removed

PR #1155, opened 2020-10-19
chore: switch to @dhis2/ui modals (DHIS2-9699)

turban:

The primary focus for this PR is to is to switch from Material UI dialogs to @dhis2/ui modals. There will still be Material UI components within the modals that will be replaced in later PRs. 

The updated components are: 
- Layer dialog (add/edit layer)
- Layer download dialog
- Open as map dialog

There are also som updated translations that originates in a previous PR #1147 

Layer modal after this PR (@dhis2/ui modal and buttons): 

Screenshot 2020-10-19 at 12 10 31

Layer dialog before this PR: 

Screenshot 2020-10-19 at 12 13 08

Open as map modal after this PR (@dhis2/ui modal and buttons): 

Screenshot 2020-10-19 at 13 02 13

Open as map dialog before this PR: 

Screenshot 2020-10-19 at 13 01 48

Layer download modal after this PR (@dhis2/ui modal and buttons): 

Screenshot 2020-10-19 at 12 49 09

In "download state": 

Screenshot 2020-10-19 at 12 53 05

Showing error: 

Screenshot 2020-10-19 at 12 59 02

Layer download before PR: 

Screenshot 2020-10-19 at 12 49 27

There are issues with z-index with this PR, so it will not be merged to master but a separate branch (chore/dhis2/ui) where these issues will be fixed before merging to master. 

Screenshot 2020-10-19 at 12 31 20

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2020-09-21T08:07:40.821Z\n"
-"PO-Revision-Date: 2020-09-21T08:07:40.821Z\n"
+"POT-Creation-Date: 2020-10-19T07:59:10.931Z\n"
+"PO-Revision-Date: 2020-10-19T07:59:10.931Z\n"
 
 msgid "Maps"
 msgstr ""
@@ -140,7 +140,7 @@ msgstr ""
 msgid "Period"
 msgstr ""
 
-msgid "Organisation units"
+msgid "Organisation Units"
 msgstr ""
 
 msgid "Style"
@@ -230,7 +230,7 @@ msgstr ""
 msgid "Valid range is"
 msgstr ""
 
-msgid "Group set"
+msgid "Group Set"
 msgstr ""
 
 msgid "Remember to select the organisation unit level containing the facilities."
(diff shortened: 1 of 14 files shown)


src/components/layers/download/styles/DataDownloadDialogActions.module.css
```diff
@@ -0,0 +1,5 @@
+.btnProgress {
+    position: absolute;
+    top: -10px;
```
martinkrulltott, line 3: One could argue that this isn't a scalable way position the circular loader in the center of the button, since if the button dimensions change (in `ui`) this will be incorrectly positioned. But I think the button changing dimensions is unlikely so 🤷‍♂️

i18n/en.pot
```diff
@@ -140,7 +140,7 @@ msgstr ""
 msgid "Period"
 msgstr ""
 
-msgid "Organisation units"
+msgid "Organisation Units"
```
martinkrulltott, line 143: Just curious, are these pot string case changes caused by a previous PR? I don't see the corresponding string changing in the js files.

martinkrulltott (approved):
Two microscopic comments, but approved as is regardless 👍

cooper-joe (approved):
LGTM

i18n/en.pot
turban: In src/components/edit/BoundaryDialog.js

Screenshot 2020-10-20 at 11 07 09

martinkrulltott: Fair enough, so the js change isn't part of this PR, no worries 👍   Classic case of "en.pot lag" then 🙂

src/components/layers/download/styles/DataDownloadDialogActions.module.css
turban: Good point. I've changed to `margin: auto` so it will always center regardless of the button size. 

Screenshot 2020-10-20 at 11 21 14

PR #1168, opened 2020-10-21
chore: switch to @dhis2/ui SingleSelectField and MultiSelectField (DHIS2-9699)

turban:

The goal of this PR is to replace all MUI TextFields with @dhis2/ui SingleSelectField and MultiSelectField. I had already created a wrapper component for the MUI TextField, and I decided to keep this wrapper to simplify the code where the select fields are used. The wrapper unifies the SingleSelectField and MultiSelectField components, allows options to be created form an array of models or objects, and allows item ids be numbers although @dhis2/ui requires option values to be strings. Reusing the select field wrapper also required less code changes across the map, as the interface is the same except renaming the "data-test" property to "dataTest". 

There is an bug with the placement of the dropdown menu which I believe is caused by conflicting Material UI styles. I also get the same error with the @dhis2/ui popover component: 

Screenshot 2020-10-21 at 15 55 23

For this reason, I'm not merging this PR to `master` branch but the `chore/dhis2/ui` branch. I will check this issue again when we get rid of all MUI styles. 

Some examples of select fields after this PR: 

Screenshot 2020-10-21 at 16 00 28
Screenshot 2020-10-21 at 16 00 38
Screenshot 2020-10-21 at 16 00 51
Screenshot 2020-10-21 at 16 01 05

Before this PR: 

Screenshot 2020-10-21 at 16 02 50
Screenshot 2020-10-21 at 16 02 57
Screenshot 2020-10-21 at 16 03 05
Screenshot 2020-10-21 at 16 03 23

### DIFF
--- a/src/components/classification/Classification.js
+++ b/src/components/classification/Classification.js
@@ -31,7 +31,10 @@ const styles = {
     },
 };
 
-const classRange = range(3, 10).map(num => ({ id: num, name: num.toString() })); // 3 - 9
+const classRange = range(3, 10).map(num => ({
+    id: num.toString(),
+    name: num.toString(),
+})); // 3 - 9
 
 const Classification = ({
     method,
@@ -75,8 +78,8 @@ const Classification = ({
 };
 
 Classification.propTypes = {
-    method: PropTypes.number,
-    classes: PropTypes.number,
+    method: PropTypes.string,
+    classes: PropTypes.string,
     colorScale: PropTypes.string,
     setClassification: PropTypes.func.isRequired,
     setColorScale: PropTypes.func.isRequired,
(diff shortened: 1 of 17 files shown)


cooper-joe:
Looking good! (apart from the weird placement bug, of course 😅 ).

Can we also update the fields for Low Radius and High Radius (thematic layer style tab) to use `ui` `inputField`, or is there an issue updating those? I believe the `inputField` supports `type='number'`.

turban:
> Can we also update the fields for Low Radius and High Radius (thematic layer style tab) to use `ui` `inputField`, or is there an issue updating those? I believe the `inputField` supports `type='number'`.

Yes, will be in a separate PR.

cooper-joe:
> > Can we also update the fields for Low Radius and High Radius (thematic layer style tab) to use `ui` `inputField`, or is there an issue updating those? I believe the `inputField` supports `type='number'`.
> 
> Yes, will be in a separate PR.

Great, thanks. LGTM!

cooper-joe (approved):

PR #1228, opened 2020-11-03
chore: refactor context menu to use @dhis2/ui (DHIS2-9699)

turban:

This PR switches from MUI to @dhis2/ui for the context menu, including an overhaul of the code. 

After this PR:

Screenshot 2020-11-03 at 10 01 02

Vertical placement of the context menu is not correct, same issue as before - will hopefully be fixed when we get rid of all MUI styles. We're not merging to master before the popover placement issues are fixed. 

There are no changes in translation file, just a reordering.


cooper-joe (approved):
LGTM. We can update the icons once ui-icons gets released.

jenniferarnesen (approved):

PR #1231, opened 2020-11-04
chore: refactor color picker to use @dhis2/ui and css modules (DHIS2-9699)

turban:

This PR is a refactoring of the color picker to use popover from @dhis2/ui,  CSS modules and functional components. 
We don't have a color picker in @dhis2/ui, so we are still using the chrome picker from react-color: https://casesandberg.github.io/react-color/

After this PR:
Screenshot 2020-11-04 at 16 15 43

Screenshot 2020-11-04 at 16 15 56

Still a placement issue for the popover, - will fix the issue when all MUI styles are gone.

### DIFF
--- a/src/components/core/ColorPicker.js
+++ b/src/components/core/ColorPicker.js
@@ -13,12 +13,12 @@ const ColorPicker = ({ color, label, width, height, onChange, className }) => {
 
     return (
         <Fragment>
-            <div className={styles.colorPicker}>
+            <div className={cx(styles.colorPicker, className)}>
                 {label && <div className={styles.label}>{label}</div>}
                 <div
                     ref={anchorRef}
                     onClick={() => setIsOpen(true)}
-                    className={cx(styles.button, className)}
+                    className={styles.button}
                     style={{
                         background: color,
                         width: width || '100%',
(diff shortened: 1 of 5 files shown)


cooper-joe (approved):
LGTM. I think we could make this screen a bit easier to understand, but that is outside the scope of this PR and should be done as a separate task if it is a priority. 

Also, just for info we have a [color picker ticket](https://jira.dhis2.org/browse/TECH-311) open for `ui`, I am pushing for it to be prioritized. A 'color picker input' will also be provided that we can use in the future.

src/components/core/ColorPicker.js
```diff
@@ -1,81 +1,62 @@
...
                     }}
-                    disableTouchRipple={true}
                 >
-                    <ArrowDropDownIcon
-                        htmlColor={hcl(color).l < 70 ? '#fff' : '#333'}
+                    <span
                         className={styles.icon}
-                    />
-                </IconButton>
+                        style={{ color: hcl(color).l < 70 ? '#fff' : '#333' }}
```
jenniferarnesen, line 30: #fff is available as `white` from ui/colors.

jenniferarnesen (approved):

src/components/core/ColorPicker.js
turban: Changed to:

```
style={{
  color: `var(--colors-${hcl(color).l < 70 ? 'white' : 'grey900'})`,
}}
```

PR #1234, opened 2020-11-05
chore: refactor layer cards to use @dhis2/ui and css modules (DHIS2-9699)

turban:

This PR is a refactoring of the layer cards in the left drawer to use @dhis2/ui, CSS modules. There is now more code sharing between the basemap card and the overlay cards above. 

The drawer and the layer toolbar still uses MUI components, and will be replaced in a later PRs. 

After this PR:

Screenshot 2020-11-05 at 16 55 31


cooper-joe (approved):

src/components/layers/basemaps/BasemapCard.js
```diff
@@ -75,61 +22,29 @@ const BasemapCard = props => {
...
-                            className={classes.expand}
-                            onClick={toggleBasemapExpand}
-                            style={{ backgroundColor: 'transparent' }}
-                        >
-                            {isExpanded ? <CollapseIcon /> : <ExpandIcon />}
-                        </IconButton>
-                    </Tooltip>
-                }
+        <LayerCard
+            title={i18n.t(name)}
```
jenniferarnesen, line 29: Does this translate correctly?

src/components/layers/overlays/OverlayCard.js
```diff
@@ -0,0 +1,111 @@
...
+    toggleLayerExpand,
+    toggleLayerVisibility,
+} from '../../../actions/layers';
+import { setMessage } from '../../../actions/message';
+import { toggleDataTable } from '../../../actions/dataTable';
+import { openDataDownloadDialog } from '../../../actions/dataDownload';
+import { setAnalyticalObjectAndSwitchApp } from '../../../util/analyticalObject';
+import styles from './styles/OverlayCard.module.css';
+
+const downloadableLayerTypes = ['facility', 'thematic', 'boundary', 'event'];
```
jenniferarnesen, line 20: What about having a layerTypes constants file that exports all of these?

turban: layer types constants are now added to contstants/layers.js

src/components/layers/basemaps/BasemapCard.js
turban: I removed `i18n.t(...)`, as translations are already handled in constants/basemaps.js and in the Maintenance app (external basemaps).

PR #1459, opened 2021-02-05
fix: improved mobile handling (DHIS2-10413)

turban:
Partly fixes: https://jira.dhis2.org/browse/DHIS2-10413 and https://jira.dhis2.org/browse/DHIS2-10512

Dependant on: https://github.com/dhis2/maps-gl/pull/276

This PR changes the fullscreen/resize handling for dashboard/plugin maps.

Logic: 
- When the user is entering fullscreen, we leave the map view as it is. When the user is leaving fullscreen, we set the map view to cover the combined layer bounds (DHIS2-10512)
- By default scroll wheel zoom is disabled for plugin maps, if entering fullscreen, we enable it again. Back to disabled when leaving fullscreen. 
- By default multiTouch is enabled for plugin maps (i.e. the maps needs to be panned by two fingers to avoid a page scroll trap). When entering fullscreen, we disable multi touch to allow "normal" touch interaction. When leaving multi touch is enabled again. 

The PR also includes a popup style fix: 

Before this PR:
Screenshot 2021-02-05 at 10 49 25

After this PR: 
Screenshot 2021-02-05 at 10 48 04


turban:
Upgraded maps-gl dependency after https://github.com/dhis2/maps-gl/pull/276 was merged.

src/util/map.js
```diff
@@ -77,3 +77,19 @@ export const drillUpDown = (layerConfig, parentId, parentGraph, level) => ({
...
 });
+
+// Called when plugin maps enter or exit fullscreen
+export const onFullscreenChange = (map, isFullscreen = false) => {
+    map.resize();
+
+    if (!isFullscreen) {
+        const bounds = map.getLayersBounds();
+
+        if (Array.isArray(bounds)) {
```
edoardo, line 88: Does it make a difference if the array is empty?

edoardo (approved):

src/util/map.js
turban: We have a check for this in `maps-gl`, so should not be needed here:
https://github.com/dhis2/maps-gl/blob/master/src/utils/geometry.js#L18

PR #1520, opened 2021-03-01
fix: keep palette and support decimals for EE layers (DHIS2-10582)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-10582

From the monkey testing: 
- Only show legend preview when the values are valid
- Min, max and step should not reset each other
- Allow steps to be changed with the keyboard
- Number of steps align the the number of colors shown, not the number of steps between min and max

In addition: 
- Show help text when values are missing or not correct

This PR upgrades DHIS2 UI to latest version 6.5.0 - with support for min, max and step values for input fields. This update also solved an issue where the color palette where changed back to default when min, max and steps where changed. 

After this PR: 

ezgif com-gif-maker

After this PR when number of steps is reduced the palette is kept: 

Screenshot 2021-03-01 at 13 37 19

Before this PR the palette was changed back to default: 

Screenshot 2021-03-01 at 13 38 34

The PR also allows for decimals precision in the classes/legend when the numbers are small: 

Screenshot 2021-03-01 at 13 43 44

Before this PR: 

Screenshot 2021-03-01 at 13 43 51

I also changed in i18n translation as we got an error then the key was including a colon character.

### DIFF
--- a/src/components/edit/earthEngine/StyleSelect.js
+++ b/src/components/edit/earthEngine/StyleSelect.js
@@ -48,7 +48,9 @@ const StyleSelect = ({ unit, params, setParams }) => {
     return (
         <div className={styles.flexColumnFlow}>
             <div className={styles.flexColumn}>
-                <p>{i18n.t('Unit: {{ unit }}', { unit })}</p>
+                <p>
+                    {i18n.t('Unit')}: {unit}
+                </p>
                 <div key="minmax" className={styles.flexInnerColumnFlow}>
                     <NumberField
                         label={i18n.t('Min')}
--- a/src/components/edit/earthEngine/StyleSelect.js
+++ b/src/components/edit/earthEngine/StyleSelect.js
@@ -14,13 +14,13 @@ import styles from '../styles/LayerDialog.module.css';
 const minSteps = 3;
 const maxSteps = 9;
 
-const isValidParams = ({ min, max }) =>
+const paramsAreValid = ({ min, max }) =>
     !Number.isNaN(min) && !Number.isNaN(max) && max > min;
 
 const StyleSelect = ({ unit, params, setParams }) => {
     const { min, max, palette } = params;
     const [steps, setSteps] = useState(palette.split(',').length);
-    const legend = isValidParams(params) && createLegend(params);
+    const legend = paramsAreValid(params) && createLegend(params);
 
     const onStepsChange = useCallback(
         steps => {
@@ -45,7 +45,7 @@ const StyleSelect = ({ unit, params, setParams }) => {
     } else if (Number.isNaN(max)) {
         warningText = i18n.t('Max value is required');
     } else if (max <= min) {
-        warningText = i18n.t('Max should be higher than min');
+        warningText = i18n.t('Max should be greater than min');
     } else if (steps < minSteps || steps > maxSteps) {
         warningText = i18n.t('Valid steps are {{minSteps}} to {{maxSteps}}', {
             minSteps,
--- a/src/components/edit/earthEngine/StyleSelect.js
+++ b/src/components/edit/earthEngine/StyleSelect.js
@@ -11,97 +11,104 @@ import { setParams } from '../../../actions/layerEdit';
 import { EE_BUFFER } from '../../../constants/layers';
 import styles from '../styles/LayerDialog.module.css';
 
-const getStepsFromParams = ({ palette, min }) =>
-    palette.split(',').length - (min === 0 ? 1 : 2);
+const minSteps = 3;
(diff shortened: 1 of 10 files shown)


src/components/edit/earthEngine/StyleSelect.js
```diff
@@ -11,40 +11,48 @@ import { setParams } from '../../../actions/layerEdit';
...
-    palette.split(',').length - (min === 0 ? 1 : 2);
+const minSteps = 3;
+const maxSteps = 9;
 
-const isValidSteps = (newSteps, params) => {
-    const steps =
-        newSteps !== undefined ? newSteps : getStepsFromParams(params);
-    return steps > 0 && steps < 8; // Valid steps: 1-7
-};
+const isValidParams = ({ min, max }) =>
```
martinkrulltott, line 17: `params` are plural so should be `areValidParams`? or `paramsAreValid`? Thoughts from a native speaker like @jenniferarnesen ?

src/components/edit/earthEngine/StyleSelect.js
```diff
@@ -11,40 +11,48 @@ import { setParams } from '../../../actions/layerEdit';
...
-                setSteps(steps);
             }
+
+            setSteps(steps);
         },
-        [params, setParams]
+        [palette, setParams]
     );
 
+    let help;
```
martinkrulltott, line 41:
```suggestion
    let warningText;
```

src/components/edit/earthEngine/StyleSelect.js
```diff
@@ -11,40 +11,48 @@ import { setParams } from '../../../actions/layerEdit';
...
+
+    if (Number.isNaN(min)) {
+        help = i18n.t('Min value is required');
+    } else if (Number.isNaN(max)) {
+        help = i18n.t('Max value is required');
+    } else if (max <= min) {
+        help = i18n.t('Max should be higher than min');
```
martinkrulltott, line 48: `higher` -> `larger`, `greater`, `bigger`? Another grammar question for the native speakers 🙂  @jenniferarnesen

martinkrulltott: Since the unit depends on the layer (e.g. elevation uses meters and population uses people per hectare) I guess we should use the most generic phrase here unless we want to swap it out based on the layer (which is a bit overkill IMHO).

martinkrulltott (commented):
Generally looks good. Added some minor comments above.
The second screenshot in the description shows steps set to 4 but 5 steps still being visible though. Correct?
image

src/components/edit/earthEngine/StyleSelect.js
turban: Changed

src/components/edit/earthEngine/StyleSelect.js
jenniferarnesen: Good question! I don't think being a native speaker makes me more of an authority on this though (but please do ask me about when to use which vs that if you want a pedantic answer!)

I don't have a strong opinion on it - maybe it feels more natural to say "paramsAreValid" when reading the code aloud to yourself.

turban:
> The second screenshot in the description shows steps set to 4 but 5 steps still being visible though. Correct?

The screenshot was created before I changed the step handling. It looks like this: 

Screenshot 2021-03-04 at 10 07 01

jenniferarnesen:
I see use of both "metres" and "meters" on the same page.  We should stick with one (not sure if we use US or UK as standard).

src/components/edit/earthEngine/StyleSelect.js
jenniferarnesen: Hmm, maybe greater?

martinkrulltott: Joe replied: 

> generally speaking, I would say greater, that’s the “mathematical” comparison term. I would only use bigger or larger if the numbers referred to something physical (a surface area etc.). so easiest to go with greater

src/components/edit/earthEngine/StyleSelect.js
turban: Changed to "paramsAreValid"

src/components/edit/earthEngine/StyleSelect.js
turban: Changed to "Max should be greater than min"

turban:
> I see use of both "metres" and "meters" on the same page. We should stick with one (not sure if we use US or UK as standard).

Changed to "meters"

martinkrulltott (approved):
LGTM 🥇

PR #1527, opened 2021-03-02
fix: wider select fields in EE data tab

turban:
This PR makes the select fields in the EE data tab full width. 

After this PR: 
Screenshot 2021-03-02 at 11 12 35

Before: 
Screenshot 2021-03-02 at 11 13 10


martinkrulltott (approved):
Looks good. Did you decide if the fields should be restricted to remain 1 row (or 2 rows) when more items are used to prevent the modal from overflowing and scrolling?

turban:
> Did you decide if the fields should be restricted to remain 1 row (or 2 rows) when more items are used to prevent the modal from overflowing and scrolling?

I'm using the MultiSelectField for DHIS2 UI - do you know if this is possible?

PR #1529, opened 2021-03-02
fix: update period in legend and include year

turban:
This PR will make sure the period name is updated when changed in the EE layer dialog. It will also include year when a weekly period is selected.

After this PR: 

Screenshot 2021-03-02 at 20 59 55

Before: 

Screenshot 2021-03-02 at 20 59 24

### DIFF
--- a/src/loaders/earthEngineLoader.js
+++ b/src/loaders/earthEngineLoader.js
@@ -1,6 +1,6 @@
 import { getInstance as getD2 } from 'd2';
 import { getEarthEngineLayer } from '../constants/earthEngine';
-import { getPeriodFromFilter } from '../util/earthEngine';
+import { getPeriodNameFromFilter } from '../util/earthEngine';
 import { getOrgUnitsFromRows } from '../util/analytics';
 import { getDisplayProperty } from '../util/helpers';
 import { toGeoJson } from '../util/map';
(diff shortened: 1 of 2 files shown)


src/util/earthEngine.js
```diff
@@ -28,6 +28,18 @@ export const getPeriodFromFilter = (filter = []) => {
...
+// Returns period name from filter
+export const getPeriodNameFromFilter = filter => {
+    const period = getPeriodFromFilter(filter);
+
+    if (!period) {
+        return null;
+    }
+
+    const { name, year = '' } = period;
+    return `${name} ${year}`;
```
jenniferarnesen, line 40: Perhaps do a whitespace trim (i.e., when year is empty string)

src/loaders/earthEngineLoader.js
```diff
@@ -79,13 +79,12 @@ const earthEngineLoader = async config => {
...
     layer.legend = {
         title: name,
-        period: period ? period.name : null,
+        period,
         groups,
         unit,
         description,
         source,
         sourceUrl,
-        ...layer.legend,
```
jenniferarnesen, line 88: Is the removal of this related to the fix?

turban: Yes, it should not be there, and was overwriting the new period name.

src/util/earthEngine.js
turban: I'll fix

turban: Changed to:

```
const { name, year } = period;
return `${name}${year ? ` ${year}` : ''}`;
```

jenniferarnesen (approved):

martinkrulltott (approved):

PR #1535, opened 2021-03-04
fix: restrict aggregation types shown

turban:
This PR restrict the aggregation types show for different Earth Engine layers, so we only show those that are relevant

"Count" (number of cells) was removed for all layers. "Sum" was removed from all layers except the two population layers. 

After this PR: 

Screenshot 2021-03-04 at 13 56 45

Before: 

Screenshot 2021-03-04 at 13 57 34

The PR also includes a fix so we don't show unit in the popup when only "sum" is selected (confusing for the user). 

After this PR: 

Screenshot 2021-03-04 at 13 51 37

Before: 

Screenshot 2021-03-04 at 13 53 29

### DIFF
--- a/src/constants/earthEngine.js
+++ b/src/constants/earthEngine.js
@@ -218,6 +218,7 @@ export const earthEngineLayers = () => [
         sourceUrl:
             'https://explorer.earthengine.google.com/#detail/USGS%2FSRTMGL1_003',
         img: 'images/elevation.png',
+        aggregation: ['min', 'max', 'mean', 'median', 'stdDev', 'variance'],
         defaultAggregation: ['min', 'max', 'mean'],
         band: 'elevation',
         params: {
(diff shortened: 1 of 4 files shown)


src/constants/earthEngine.js
```diff
@@ -262,7 +264,8 @@ export const earthEngineLayers = () => [
         sourceUrl:
             'https://explorer.earthengine.google.com/#detail/MODIS%2FMOD11A2',
         img: 'images/temperature.png',
-        defaultAggregation: ['min', 'max', 'mean'],
+        aggregations: ['min', 'max', 'mean', 'median', 'stdDev', 'variance'],
```
jenniferarnesen, line 267: Might be useful to have all these strings as constants somewhere in the app?  (an idea for future refactoring)

jenniferarnesen (approved):

PR #1537, opened 2021-03-05
fix: show updated EE aggregations in data table and download

turban:
This PR solves various issues with EE aggregations in data table and download: 
- EE data tables can now be filtered, although it takes some time as new calculations needs to be performed on the filtered data. I've created a separate issue to speed up this process: https://jira.dhis2.org/browse/DHIS2-10654 (probably 2.37)
- The data table should now be updated every time new aggregations values are available. 
- Classed aggregations (i.e. landcover) will now show in the data table. Will be cramped on smaller screens, a separate issue is created to fix it: https://jira.dhis2.org/browse/DHIS2-10655 (probably 2.37)
- Geometry type was added to EE data table

This issue was fixed by adding an "aggregations" entry to the redux store which can hold aggregations values for multiple layers using the layer id as the key. Whenever this store is updated, changes are propagated to the data table, and available from the download dialog. 

After this PR: 

Screenshot 2021-03-08 at 10 52 23

ezgif com-gif-maker (1)

Screenshot 2021-03-08 at 11 05 49
(in data download)


janhenrikoverland:
Temperature layer on Sierra Leone districts -> show data table -> add a filter, e.g. min temperature `>20` -> I get data for the whole world, not just Sierra Leone

janhenrikoverland:
2. When I add the `>20` filter, the filtering fires immediately on the first character. So it tries to filter on `>` only which then leads to a red error notification saying "Parameter 'collection' is required". This filter request, which is potentially very heavy, should as a minimum be debounced, maybe even validated before anything is fired @cooper-joe

janhenrikoverland:
3. When I remove the `>20` filter I get an error message:
Screenshot from 2021-03-09 09-25-42

cooper-joe:
> This filter request, which is potentially very heavy, should as a minimum be debounced, maybe even validated before anything is fired @cooper-joe

Agreed. Debounce as a minimum. Validation is nice, if it can be done effectively with ~0% of a false positive (i.e. the user should never be blocked because our validation check is wrong).

turban:
> 1. Temperature layer on Sierra Leone districts -> show data table -> add a filter, e.g. min temperature `>20` -> I get data for the whole world, I would expect to only get Sierra Leone here as well.

Fixed.

Screenshot 2021-03-09 at 14 11 25

turban:
> 1. When I add the `>20` filter, the filtering fires immediately on the first character. So it tries to filter on `>` only which then leads to a red error notification saying "Parameter 'collection' is required". This filter request, which is potentially very heavy, should as a minimum be debounced, maybe even validated before anything is fired @cooper-joe

I've added a 500ms debounce, and fixed the error occurring when we passed in an empty feature collection.

turban:
> 1. When I remove the `>20` filter I get an error message:

Fixed here: https://github.com/dhis2/maps-gl/pull/294

I'll upgrade maps-gl in this PR when https://github.com/dhis2/maps-gl/pull/294 is approved and merged.

turban:
@janhenrikoverland and @cooper-joe: data table filtering for Earth Engine layers should work with the latest changes, but it's not optimal as the calculations happening on the EE servers are slow. 

We could consider to remove support for data table filtering for EE layers in 2.36 - this is the first time we have data tables for EE layers, so we're not removing a feature from the user.

turban:
Setting draft status - we need a better solution where we don't run calculations when the data table is filtered.

turban:
I took a second round and it is now possible to filter EE layers in a more performant way: https://github.com/dhis2/maps-gl/pull/295

When the maps-gl PR is approved and merged, I will update this PR so it's possible to test the branch. 

After this PR: 


The default legend for temperature and precipitation is also improved:

Screenshot 2021-03-09 at 21 16 59

Screenshot 2021-03-09 at 21 16 44

turban:
dhis2/maps-gl#295 is merged and this PR can now be reviewed and tested @janhenrikoverland

janhenrikoverland (approved):

PR #1544, opened 2021-03-08
fix: show circular loader in org unit dialog

turban:
This PR shows a circular loader while data is loading in the org unit dialog: 

Screenshot 2021-03-08 at 18 07 50


janhenrikoverland (approved):
Works well

PR #1565, opened 2021-03-11
fix: upgrade dhis2 library dependencies to support tree-shaking (DHIS2-9893)

amcgee:
Part of [DHIS2-9893](https://jira.dhis2.org/browse/DHIS2-9893)

Upgrades ui, app-runtime, d2-i18n, and analytics

EDIT: I went ahead and refactored the webpack configuration here, since this was using webpack 3 and didn't support treeshaking - need to investigate a locale initialization issue, but otherwise this is good to go (and reduces bundle sizes a fair bit!)

This reduces app bundle size by about 500kb and plugin bundle size by about 350kb (both gzipped)


varl (approved):
Tested locally and I did not spot any immediate problems.

amcgee:
@varl @turban see edit, last commit is a bit bigger but makes things MUCH cleaner and supports treeshaking 1e370ba

This removes all use us DHIS_CONFIG files, instead relying on a DHIS2_BASE_URL environment variable.  Upgrades webpack and all loaders, uses the babel config from cli-app-scripts, and removes a bunch of unused dependencies.

amcgee:
If the last commit here is too disruptive we can back it out and reapply on master

turban (approved):

janhenrikoverland (approved):

PR #1685, opened 2021-05-05
feat: setOnlineStatus method for map plugin to control context menu visibility (DHIS2-11053)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11053

`mapPlugin.setOnlineStatus('map', false)`

### DIFF
--- a/src/components/plugin/Plugin.js
+++ b/src/components/plugin/Plugin.js
@@ -32,6 +32,7 @@ class Plugin extends Component {
         super(props, context);
 
         this.state = {
+            isOnline: true,
             mapViews: props.mapViews, // Can be changed by drilling
             resizeCount: 0,
         };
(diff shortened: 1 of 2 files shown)


jenniferarnesen:
Here is what the spec says:

```
Drill-down / Drill-up
Drilling up and down levels in map data is not supported. The Drill up one level and Drill down one level menu items are disabled. If possible, a tooltip is displayed on hover with the content “Not available offline”.
Panning, Zooming, Tilting
Moving around in a map by panning, zooming or tilting is supported, even without the ability to load new basemap data.

```

I think the button still needs a disabled state with a tooltip?

turban:
I've updated PR so it (hopefully) aligns with the design spec from @cooper-joe

PR #1720, opened 2021-05-26
feat: org unit layer (DHIS2-11071) 

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11071

This PR will: 
- re-purpose the existing boundary layer to a new org unit layer in DHIS2 Maps
- make the facility layer more flexible and easier to use


turban:
Superseded by other PRs

PR #1732, opened 2021-06-01
fix: configure cypress tests to run on CI against live instance

jenniferarnesen:
This is just the initial step for getting cypress back into the workflow for maps. Next will be to enable more tests and organize them, and to use the network shim.

The test instance is the one used by dashboard-app and automatically gets reset every other day. Shouldn't cause any issues.


amcgee (approved):
LGTM

turban (approved):

PR #1734, opened 2021-06-03
fix: use period generators from dhis2/analytics (DHIS2-11266) 

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11266

This PR will switch from d2/period/generators, which is no longer up-to-date, to the period generators in @dhis2/analytics. We also add more fixed period to align it with Data Visualizer: 
- Weekly (Start Wednesday)
- Weekly (Start Thursday)
- Weekly (Start Saturday)
- Weekly (Start Sunday)
- Bi-weekly
- Financial year (Start November)

Locale is no longer passed to the period generator as it is not supported by @dhis2/analytics (it was used in d2 to create a language sensitive date string using `toLocaleDateString`). 

This PR will also update the d2-ui dependencies to the latest version (7.3.0)

Hiding periods based on system settings will be in a separate PR: https://jira.dhis2.org/browse/DHIS2-11245

After this PR: 

Screenshot 2021-06-07 at 17 39 06

Screenshot 2021-06-07 at 17 43 30

fixedperiods


src/constants/periods.js
```diff
@@ -2,17 +2,23 @@ import i18n from '@dhis2/d2-i18n';
...
+    { id: 'QUARTERLY', name: i18n.t('Quarterly') },
+    { id: 'SIXMONTHLY', name: i18n.t('Six-monthly') },
+    { id: 'SIXMONTHLYAPR', name: i18n.t('Six-monthly April') },
+    { id: 'YEARLY', name: i18n.t('Yearly') },
+    { id: 'FYNOV', name: i18n.t('Financial year (Start November)') },
+    { id: 'FYOCT', name: i18n.t('Financial year (Start October)') },
+    { id: 'FYJUL', name: i18n.t('Financial year (Start July)') },
+    { id: 'FYAPR', name: i18n.t('Financial year (Start April)') },
     { id: 'StartEndDates', name: i18n.t('Start/end dates') },
 ];
```
janhenrikoverland, line 24: Import the ids from here: https://github.com/dhis2/analytics/blob/7de0afe0b16f719cd0949d2f84487de18f4c4871/src/index.js#L233 ?

turban: It increases the code complexity, but will change so we follow the same code pattern across the apps.

turban: Fixed.

janhenrikoverland (approved):
LG

PR #1745, opened 2021-06-09
feat: hide periods based on system settings (DHIS2-11245)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11245

This PR will hide some of the fixed and relative periods based on system settings. I have tried to reuse code from @dhis2/analytics where possible (available for import). 

Notable changes: 
- Read hidden periods from system settings
- Hide fixed and relative periods based on above settings
- Make sure default relative period is not hidden before selecting it
- Fallback to start/end dates for event layer if 1) no default relative period is set or 2) it is hidden. I have not done the same for thematic layer, as we can not be sure if data is aggregated over daily periods. 
- Fetch all relative periods from @dhis2/analytics, instead of defining locally for DHIS2 Maps. 
- Use period constants instead of string ids - constants are read from @dhis2/analytics if available.

After this PR:

If default relative period is set, and none are hidden: 

Screenshot 2021-06-10 at 12 43 19

All periods will show in period selects: 

Screenshot 2021-06-10 at 13 04 28

Screenshot 2021-06-10 at 13 04 50

The default relative period will be selected by default for thematic layer: 

Screenshot 2021-06-10 at 12 45 07

and for event layer: 

Screenshot 2021-06-10 at 12 47 17

If the default period is by month while monthly periods are hidden:

Screenshot 2021-06-10 at 12 49 42

Monthly periods will not show for fixed periods:

Screenshot 2021-06-10 at 13 06 47

and for relative periods: 

Screenshot 2021-06-10 at 13 07 08

No period will be selected for thematic layers: 

Screenshot 2021-06-10 at 12 52 02

For event layers the fallback will be on default start and end dates (last year): 

Screenshot 2021-06-10 at 12 53 08


janhenrikoverland (approved):
LG

PR #1759, opened 2021-06-14
fix: use hyphen in locale strings passed to the browser(DHIS2-11335)

turban:
Fixes for 2.35: https://jira.dhis2.org/browse/DHIS2-11335

DHIS2 uses underscore to define locales, while the browser/JS APIs use hyphen (BCP 47). This PR will replace underscores with hyphens before the locale is passed to ECMAScript Internationalization API. 

Avoids a crash when opening the interpretations panel with a locale with language_country format
See dhis2/d2-ui#700

After this PR with interface language set to "Arabic (Iraq)": 
Screenshot 2021-06-24 at 13 41 50

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
         "@dhis2/d2-ui-analytics": "^0.0.3",
         "@dhis2/d2-ui-core": "^7.1.6",
         "@dhis2/d2-ui-file-menu": "^7.1.6",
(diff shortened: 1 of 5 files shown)


turban:
Set to draft status until we have a fix for the same issue in @dhis2/d2-ui-interpretations

edoardo (approved):

PR #1764, opened 2021-06-17
fix: exclude attribution toggle button from download (DHIS2-11143)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11143

2.34 backport of #1712

This PR will exclude the attribution toggle button from the map download, as it's not supported by the dom-to-image dependency. It will show the text attribution instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "maps-app",
-    "version": "34.0.47",
+    "version": "34.0.48",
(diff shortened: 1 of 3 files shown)


turban:
After this PR:
map-download

Before: 
Screenshot 2021-06-17 at 12 28 52

jenniferarnesen (approved):

PR #1833, opened 2021-08-31
fix: dhis2 dependencies upgrade

turban:
This PR will update the following dependencies to the latest version: 
- @dhis2/analytics
- @dhis2/app-runtime
- @dhis2/cli-app-scripts
- @dhis2/ui

It required a new portal root node to correct the layer ordering for UI modals. It can be removed again when DHIS2 Maps is ported to the app platform. 

Some snapshot tests also required some small adjustments. 

After this PR, the new fil open dialog is shown: 
Screenshot 2021-09-01 at 13 52 29

Before: 
Screenshot 2021-09-01 at 13 53 15
 
Will fix cypress test in a separate PR.


jenniferarnesen:
@turban The cypress tests are actually catching a real bug in this case. Looks like a z-index problem.

image

edoardo (approved):

PR #1837, opened 2021-09-02
fix: organisation unit color

turban:
Partly fixes: https://jira.dhis2.org/browse/DHIS2-11071

This PR will make it possible to set the organisation unit color for facility and org unit layers. Backend support was just added.

Screenshot 2021-09-02 at 14 12 06

Screenshot 2021-09-02 at 14 11 28

After this PR: 
Screenshot 2021-09-02 at 13 50 17

Screenshot 2021-09-02 at 13 50 58

Before: 
Screenshot 2021-09-02 at 13 52 05

Screenshot 2021-09-02 at 13 52 41


martinkrulltott (approved):
LGTM. I see a vivid mix of "orgUnit" and "organisationUnit" in the code though, maybe it would be best to stick to one or the other, but I suppose that refactoring is way out of scope for this PR.

PR #1847, opened 2021-09-09
fix: check if response body is empty

turban:
This PR fixes an issue with new configuration endpoints in 2.37: 
https://debug.dhis2.org/dev/api/37/configuration/facilityOrgUnitLevel
https://debug.dhis2.org/dev/api/37/configuration/facilityOrgUnitGroupSet

If the config is not set, it returns an empty body, which gives an error when we use response.json()

This fix checks if the response is an empty string before trying to parse it. 
I found the solution on https://stackoverflow.com/a/51320025

@larshelge is this empty response by design?

### DIFF
--- a/src/util/api.js
+++ b/src/util/api.js
@@ -1,12 +1,12 @@
 import { isString, isObject } from 'lodash/fp';
 import { config } from 'd2';
 
-// The api/configurations/xx endpoints returns an empty string if no config is set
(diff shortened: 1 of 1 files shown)


turban:
Reply form Lars: It is like this for all configuration endpoints. We need to live with it until further notice.

jenniferarnesen (approved):

PR #1856, opened 2021-09-10
fix: facility layer and group set styling improvements

turban:
This PR fixes the following issues for facility layer: 
- Upgrades maps-gl dependency with symbol opacity fix https://github.com/dhis2/maps-gl/pull/395
- Extra space above the group set preview 
- More bullet proof fallback symbols and colors
- Made the fallback colors more distinguishable

Colors and icon symbols for org unit groups sets needs to be set the maintenance app, but we support some simple fallbacks.

For icon symbols the fallback is colored circles. The icon set we have used for a long time only contains 5 color circles. This PR makes sure we select one of those. If there are more than 5 groups without symbols, the same color circle is used. 

Same with colors except that we have unlimited number of colors: If a group is not having a color set, we will pick from our 12 predefined colors, or a random color if there are more groups.

The icon set should be improved - and we should also allow users to upload their own icons - but it won't happen for 2.37.  

After this PR (more distinguishable colors + extra space above legend): 

Screenshot 2021-09-10 at 13 24 58

Before: 

Screenshot 2021-09-10 at 13 25 20

After this PR: 

Screenshot 2021-09-10 at 13 28 03

Before: 

Screenshot 2021-09-10 at 13 28 27

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
         "@dhis2/d2-ui-interpretations": "^7.3.2",
         "@dhis2/d2-ui-org-unit-dialog": "^7.3.2",
         "@dhis2/d2-ui-org-unit-tree": "^7.3.2",
-        "@dhis2/maps-gl": "^2.1.0",
+        "@dhis2/maps-gl": "^2.1.1",
         "@dhis2/ui": "^6.20.0",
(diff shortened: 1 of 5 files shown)


martinkrulltott (commented):
LGTM

PR #1864, opened 2021-09-13
fix: remove relocate facility options from context menu (DHIS2-9957)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-9957

"Swap longitude/latitude" and "Relocate" is no longer supported by the backend API, and this PR will remove the options from the context menu. 

It is not likely that we will add these options to the maps app again, but it should be possible in the maintenance app. 

After this PR: 

Screenshot 2021-09-13 at 15 43 41

Before:  

Screenshot 2021-09-13 at 15 41 08


src/components/map/ContextMenu.js
```diff
@@ -39,39 +32,21 @@ const ContextMenu = (props, context) => {
...
-    if (feature) {
-        const { geometry, properties } = feature;
-
-        attr = properties || {};
-
-        // Treat bubbles as polygons if created from one
-        isPoint =
-            geometry.type === 'Point' && !polygonTypes.includes(attr.type);
-    }
+    const attr = feature ? feature.properties : {};
```
janhenrikoverland, line 49: `feature?.properties || {}`

janhenrikoverland (approved):
Approved with comment

src/components/map/ContextMenu.js
turban: Fixed.

janhenrikoverland (approved):

PR #1896, opened 2021-10-02
fix: keep start/end dates for event layer (DHIS2-11919)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11919

This PR makes sure that the default start/end dates are only set once. 

After this PR the date in the layer dialog is the same as layer card: 
Screenshot 2021-10-02 at 17 49 56

Before the dates are noe equal: 
Screenshot 2021-10-02 at 17 50 21

### DIFF
--- a/src/components/edit/event/EventDialog.js
+++ b/src/components/edit/event/EventDialog.js
@@ -142,8 +142,7 @@ export class EventDialog extends Component {
             setPeriod({
                 id: defaultPeriod,
             });
-        } else {
-            // Fallback to default start/end dates
+        } else if (!startDate && !endDate) {
             setStartDate(DEFAULT_START_DATE);
             setEndDate(DEFAULT_END_DATE);
         }


src/components/edit/event/EventDialog.js
```diff
@@ -142,8 +142,7 @@ export class EventDialog extends Component {
             setPeriod({
                 id: defaultPeriod,
             });
-        } else {
-            // Fallback to default start/end dates
+        } else if (!startDate && !endDate) {
```
jenniferarnesen, line 145: Is the "else" case handled correctly? (if there is a plausible "else" case)

jenniferarnesen (approved):

src/components/edit/event/EventDialog.js
turban: It is no plausible else case here - it means that a period is already set - and we should not apply a default.

PR #1929, opened 2021-11-01
fix: use native input instead of dhis2/ui input in order to correctly validate manually entered dates

jenniferarnesen:
Backport of https://github.com/dhis2/maps-app/pull/1925

Fixes https://jira.dhis2.org/browse/DHIS2-11934

Controlled input type date components are problematic because of the way date validation is handled by the native element.

Use an uncontrolled input component, pass in a default value, and use onBlur instead of onChange to set the date value in the redux store.

Styles are copied from dhis2/ui input.


turban (approved):
LGTM

PR #1930, opened 2021-11-03
fix: maps-gl upgrade with earth engine fixes

turban:
Improves: https://jira.dhis2.org/browse/DHIS2-12013

The maps-gl upgrade include two small fixes that are already approved:

**Don't update auth library for earth engine api**
We don't need to update the Google authentication library when we set the token for the EE API. This small change will remove two Google requests. https://github.com/dhis2/maps-gl/pull/414

**Only reference image scale on the server**
The image scale value is only used in server processing, and we don't need to get the value on the client side. https://github.com/dhis2/maps-gl/pull/416

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
         "@dhis2/d2-ui-interpretations": "^7.3.3",
         "@dhis2/d2-ui-org-unit-dialog": "^7.3.3",
         "@dhis2/d2-ui-org-unit-tree": "^7.3.3",
-        "@dhis2/maps-gl": "^2.2.1",
+        "@dhis2/maps-gl": "^2.2.3",
         "@dhis2/ui": "^6.20.0",
         "abortcontroller-polyfill": "^1.5.0",
         "array-move": "^3.0.1",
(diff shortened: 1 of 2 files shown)


janhenrikoverland (approved):

turban:
@jenniferarnesen, not merging this PR as a Cypress test is failing. Do you see why it is happening? 

When I try to run the test the Cypress tests locally I get another error: 

Screenshot 2021-11-03 at 11 10 42

I've added "cypress.env.json" to the root folder with `dhis2_base_url": "http://localhost:8080"`.

jenniferarnesen:
> @jenniferarnesen, not merging this PR as a Cypress test is failing. Do you see why it is happening?
> 
> When I try to run the test the Cypress tests locally I get another error:
> 
> Screenshot 2021-11-03 at 11 10 42
> 
> I've added "cypress.env.json" to the root folder with `dhis2_base_url": "http://localhost:8080"`.

The tests passed on CI now. Are you still having this problem?

PR #1937, opened 2021-11-22
fix: parse date string to array to get correct date in all timezones

jenniferarnesen:
fixes: https://jira.dhis2.org/browse/DHIS2-12169

If a string with this format "2018-10-01" is passed to javascript Date, the assumed time is "00:00" and the returned date will be according to the user's timezone. So if you are in <= GMT-1, then the resulting date will actually be the previous day. To avoid this, the date string is converted to an array, and Date converts this correctly.

The screenshot shows the Layer card with the incorrect date shown (note that the analytics request uses the correct date)

![image](https://user-images.githubusercontent.com/6113918/142909755-25bb0b87-7250-4bfa-8855-b51773bcb316.png
Fixed:

https://user-images.githubusercontent.com/6113918/142910473-5506abaa-ddb2-48e4-bbaa-6e996497e007.mov

### DIFF
--- a/src/loaders/eventLoader.js
+++ b/src/loaders/eventLoader.js
@@ -21,7 +21,7 @@ import {
     EVENT_COLOR,
     EVENT_RADIUS,
 } from '../constants/layers';
-import { formatLocaleDate } from '../util/time';
+import { formatStartEndDate, getDateArray } from '../util/time';
 import { cssColor, getContrastColor } from '../util/colors';
 
 // Server clustering if more than 2000 events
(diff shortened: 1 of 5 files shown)


turban (approved):
Tested affected layers locally, and it seems to work fine! Do you see why the cypress tests are failing?

jenniferarnesen:
> Tested affected layers locally, and it seems to work fine! Do you see why the cypress tests are failing?

The one cypress check shows failed due to a previous failed run, and I don't know why it doesn't get updated when the tests were re-run and passed.

PR #1970, opened 2021-12-08
Show aggregate values for each age group (DHIS2-11039)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-11039

Depends on this PR in maps-gl: https://github.com/dhis2/maps-gl/pull/422
Will be merged to the EE web worker refactor PR in maps-app: https://github.com/dhis2/maps-app/pull/1931

This PR will show aggregation values for each age/gender groups for the population layer. Currently we only show the aggregated values for all age/gender groups combined.

After this PR:
Screenshot 2021-12-08 at 17 55 09

Before: 
Screenshot 2021-12-08 at 17 56 11

If only one group is selected after this PR:
Screenshot 2021-12-08 at 18 02 43

Before: 
Screenshot 2021-12-08 at 18 03 54

This PR also adds a restriction to select max 10 groups (selecting more will result in very long response time from EE API):
Screenshot 2021-12-08 at 18 06 32

Aggregated values for each age/gender group will not show in the data table, as there is no room with the current layout: 
Screenshot 2021-12-08 at 18 22 39

Support can be added by having expandable rows - which is included in: https://jira.dhis2.org/browse/DHIS2-10655

All aggregation values are included when downloading data.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-12-08T15:04:35.136Z\n"
-"PO-Revision-Date: 2021-12-08T15:04:35.136Z\n"
+"POT-Creation-Date: 2021-12-08T15:23:58.705Z\n"
+"PO-Revision-Date: 2021-12-08T15:23:58.705Z\n"
 
 msgid "Maps"
 msgstr ""
@@ -180,6 +180,9 @@ msgid "Add layer"
 msgstr ""
 
 msgid "Aggregation method"
+msgstr ""
+
+msgid "Maximum number of groups are {{maxBands}}"
 msgstr ""
 
 msgid "Groups"
@@ -1304,8 +1307,4 @@ msgid "End date is invalid"
 msgstr ""
 
 msgid "End date cannot be earlier than start date"
-msgstr ""
-
-msgctxt "BANDS}}"
-msgid "Maximum number of groups are {{MAX"
 msgstr ""
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-12-10T16:33:24.098Z\n"
-"PO-Revision-Date: 2021-12-10T16:33:24.098Z\n"
+"POT-Creation-Date: 2021-12-13T12:22:02.535Z\n"
+"PO-Revision-Date: 2021-12-13T12:22:02.535Z\n"
 
 msgid "Maps"
 msgstr ""
@@ -180,6 +180,9 @@ msgid "Add layer"
 msgstr ""
 
 msgid "Aggregation method"
+msgstr ""
+
+msgid "Choosing many groups takes a long time to calculate and display."
 msgstr ""
 
 msgid "Groups"
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-11-16T09:45:00.428Z\n"
-"PO-Revision-Date: 2021-11-16T09:45:00.428Z\n"
+"POT-Creation-Date: 2021-12-08T15:04:35.136Z\n"
+"PO-Revision-Date: 2021-12-08T15:04:35.136Z\n"
 
 msgid "Maps"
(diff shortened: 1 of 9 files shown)


i18n/en.pot
```diff
@@ -182,6 +182,9 @@ msgstr ""
 msgid "Aggregation method"
 msgstr ""
 
+msgid "Maximum number of groups are {{maxBands}}"
```
cooper-joe, line 185: I suggest wording this `Earth engine layers can’t have more than {{maxBands}} groups.`.

src/components/map/layers/earthEngine/styles/EarthEnginePopup.module.css
```diff
@@ -8,6 +8,8 @@
     width: 100%;
     border-collapse: collapse;
     font-size: 14px;
```
cooper-joe, line 10: I suggest `font-size: 13px;` for the table to increase information density.

src/components/map/layers/earthEngine/styles/EarthEnginePopup.module.css
```diff
@@ -30,36 +32,46 @@
     font-weight: normal;
 }
 
-.table thead th {
-    text-align: left;
-    padding-bottom: var(--spacers-dp4);
+.table thead {
+    border-bottom: 1px solid #eee;
 }
```
cooper-joe, line 37: I suggest styling the `thead` with `color: var(--colors-grey600); font-weight: 500;`

src/components/map/layers/earthEngine/styles/EarthEnginePopup.module.css
```diff
@@ -30,36 +32,46 @@
...
 
-.table thead th:last-child {
-    text-align: right;
+.table tfoot {
+    border-top: 1px solid #eee;
```
cooper-joe, line 40: Replace `#eee` with `var(--colors-grey300)`.

src/components/map/layers/earthEngine/styles/EarthEnginePopup.module.css
```diff
@@ -30,36 +32,46 @@
...
 }
 
-.table tbody th {
-    text-align: left;
+.table tfoot {
+    border-top: 1px solid #eee;
```
cooper-joe, line 44: Replace `#eee` with `var(--colors-grey300)`.

src/components/map/layers/earthEngine/styles/EarthEnginePopup.module.css
```diff
@@ -30,36 +32,46 @@
...
+.table tbody th,
+.table tfoot th {
     text-align: left;
 }
 
-.table tbody td:last-child {
-    text-align: right;
+.table thead th,
+.table tfoot th {
+    font-weight: bold;
```
cooper-joe, line 61: Use `font-weight: 500` instead I think.

cooper-joe (commented):
I've added some design tweak suggestions. Another, generally:
- Apply `color: var(--colors-grey900);` to the whole popup.

turban:
@cooper-joe I have updated the popup styles - it now look like this: 
Screenshot 2021-12-10 at 17 32 31

The group error is changed to a warning: 
Screenshot 2021-12-10 at 17 26 38

The user can select as many groups as she likes, but it will result in slow loading times. Open for other wordings :-) 

When many groups are selected a vertical scroll will show in the popup: 
Screenshot 2021-12-10 at 17 30 26

src/components/edit/earthEngine/BandSelect.js
```diff
@@ -5,16 +5,29 @@ import { connect } from 'react-redux';
...
-    />
-);
+const MAX_BANDS = 10;
+
+const BandSelect = ({ band = [], bands, setBand, errorText }) => {
+    let warning;
+
+    if (band.length > MAX_BANDS) {
+        warning = i18n.t(
+            'Warning: It takes longer time to calculate data for many groups.'
```
cooper-joe, line 15: Does it also consume a lot of data to choose a lot of groups? If so, we could say:
`Choosing many groups downloads a lot of data and takes time to display.`
If it's just calculation time, then it could be:
`Choosing many groups takes a long time to calculate and display.`

cooper-joe:
Looks good @turban - I just added a suggestion wording for the warning. Otherwise LGTM 👍

src/components/edit/earthEngine/BandSelect.js
turban: Thanks @cooper-joe! It's just calculation time so used the latter.

src/components/edit/earthEngine/BandSelect.js
```diff
@@ -5,16 +5,29 @@ import { connect } from 'react-redux';
...
-    />
-);
+const MAX_BANDS = 10;
+
+const BandSelect = ({ band = [], bands, setBand, errorText }) => {
+    let warning;
+
+    if (band.length > MAX_BANDS) {
+        warning = i18n.t(
+            'Choosing many groups takes a long time to calculate and display.'
```
jenniferarnesen, line 15: Because calculations may take a long time, a maximum of ${MAX_BANDS} groups is allowed

turban: We have changed the logic so MAX_BANDS is no longer a restriction, but we show a warning to the user. I renamed the MAX_BANDS to WARNING_BANDS and added a code comment to make it clearer.

jenniferarnesen (approved):

PR #1984, opened 2021-12-21
feat: use default basemap that is configured in system settings [DHIS2-12155]

jenniferarnesen:
https://user-images.githubusercontent.com/6113918/147556246-53aad0be-b529-4fca-a273-bb28642db732.mov

`keyDefaultBaseMap` is fetched as part of the system settings. If no value found (there is no default value set on the server side), then the app and plugin use `osmLight` as the default. For the app, a new react hook has been added to get the defaultBasemap. For the plugin, the basemap config is added as before except it now uses `createExternalLayerConfig` for external basemaps to be inline with the app.

Refactor: All map config modifications ( `getBasemap`,`upgradeGisAppLayers` and `renameBoundaryToOrgUnit`) are moved to one module (getMigratedMapConfig.js) and run in the same order as before as soon as the map is fetched.

FileMenu: former epics (loadFavorite, saveFavorite, saveNewFavorite) have been deleted, and those functions are now in FileMenu and use the dataEngine (openMap, saveMap, saveAsNewMap). The new functions in FileMenu do the exact same work as what was previously in the epics. Errors are caught and the user is shown an alert bar (screenshot below)

MapView: added a loading spinner, otherwise no change

util/requests: `mapRequest` and `fetchMap` are identical except that mapRequest uses d2 and fetchMap uses dataEngine. I decided not to try to introduce dataEngine into the plugin yet.

Take a somewhat close look at the jest tests for the `getMigratedMapConfig` module. Especially the test configs, because I am not too familiar with the old map config formats (basemap in mapviews, GisApp layers, boundary layers), so I set up the configs based on the code. Would be good to confirm those are correct.

*****
*Some changes have been made in order to support cypress testing on the plugin:*
* scripts/writeMapCypress.js script runs on the `yarn start` command. Creates the testable plugin src/mapcypress.js which is exactly the same as map.js, but it writes to 5 test <span> elements in plugin.html:
```
document.getElementById('basemapId').textContent = config.basemap.id;
document.getElementById('basemapName').textContent = config.basemap.name;
document.getElementById('basemapIsVisible').textContent = config.basemap.isVisible === false ? 'no' : 'yes';
document.getElementById('mapviewsCount').textContent = config.mapViews.length;
document.getElementById('mapviewsNames').textContent = config.mapViews.map(view => view.layer).join(' ');
```
We need these present in order to check the map config, since we can't actually test the map itself since it is on a <canvas>
* webpack.config.js: in production mode, the plugin is created from src/map.js. In development the plugin is created from src/mapcypress.js.
*****

If saving a map fails, an alert bar now shows a message (for a 409 error, the only text back from dataEngine is "Forbidden"):
image

There were some structural changes to `src/map` that were difficult to add tests for (we need cypress screenshot tests - to be added later). So here are some videos of different types of configs sent to the plugin:

// chart config (e.g. View as Map from dashboard). system settings  keyDefaultBaseMap = osmStreetView

https://user-images.githubusercontent.com/6113918/147554930-7bae546c-0a48-4fe3-ae94-c519cbd9bb3c.mov

// map config with basemap = "none"

https://user-images.githubusercontent.com/6113918/147555176-92e26955-b66e-446b-b733-daa1457fa8a0.mov

// map config with external basemap

https://user-images.githubusercontent.com/6113918/147555538-9a6016e8-9aaf-4fe8-9583-ef1cc93b063f.mov

//map config no id, with external basemap

https://user-images.githubusercontent.com/6113918/147555842-e4fb974c-d92e-4ce6-8ebd-cae4dc6b9030.mov


src/components/layers/basemaps/BasemapCard.js
```diff
@@ -11,71 +12,70 @@ import {
...
-        >
-            <BasemapList
-                selectedID={props.basemap.id}
-                basemaps={props.basemaps}
-                selectBasemap={props.selectBasemap}
-            />
-        </LayerCard>
+        <>
+            {basemap.id === undefined ? (
+                <>
```
turban, line 32: Is fragment tag and empty div needed?

src/components/map/MapView.js
```diff
@@ -27,35 +28,48 @@ const MapView = props => {
...
-                    basemap={basemap}
-                    layer={splitViewLayer}
-                    controls={mapControls}
-                    feature={feature}
-                    openContextMenu={openContextMenu}
-                    resizeCount={resizeCount}
-                />
+        <>
+            {basemap.id === undefined ? (
+                <>
```
turban, line 33: Is fragment tag and empty div needed?

src/util/__tests__/getMigratedMapConfig.spec.js
```diff
@@ -0,0 +1,175 @@
...
+        })
+    );
+});
+
+// TODO - add a test that checks externalLayer and basemap
+test('getMigratedMapConfig with old GIS app format and Boundary layer', () => {
+    const config = {
+        id: 'mapId',
+        name: 'map name',
+        basemap: { id: 'osmStreet' },
```
jenniferarnesen, line 155: @turban Do the old configurations have a separate `basemap` property like this or is it in the mapViews (as in the first test in this file)

src/map.js
```diff
@@ -92,50 +92,52 @@ function PluginContainer() {
...
-                    },
-                };
+            if (isValidUid(config.basemap.id)) {
+                const externalLayer = await getExternalLayer(config.basemap.id);
+                basemap.config = createExternalLayerConfig(externalLayer);
             } else {
-                basemap = defaultBasemaps().find(map => map.id === basemapId);
+                basemap.config = defaultBasemaps().find(
+                    map => map.id === config.basemap.id
+                ).config;
```
turban, line 136: I'm getting an error when testing http://localhost:8082/plugin.html?id=zDP78aJU8nX

_Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'config')
    at loadLayers (map.js:135)_

There are stored maps that don't have any basemap property, example: 
https://play.dhis2.org/dev/api/37/maps/zDP78aJU8nX.json

The code will break here as config.basemap.id = undefined

When no basemap is set for a saved map, the fallback should be the default basemap - this seems to work fine in the app.

src/util/__tests__/getMigratedMapConfig.spec.js
turban: What I know: 
- basemap prop should be a string (when new maps are saved)
- basemap prop should be "none" if the basemap visibility is off when saved 
- basemap prop is not always present (fallback to default basemap)
- basemap is google maps, then fallback to equivalent bing maps
- basemap is uid, use the external layer with the same uid
- basemap is object with id, only use the id (haven't seen this lately, could be used by the old GIS app)
- if basemap id is unkown (none of above), fallback to default basemap

For new maps, an external basemap is saved by storing the basemap uid in the basemap prop. External overlays are stored as a separate layer under mapViews. Previously also external basemaps were stored under mapViews, and there are still some saved maps using this format that we should still support. Try: "Malaria cases Sierra Leone terrain basemap".

src/map.js
turban: I'm getting the same error is basemap is "none"

jenniferarnesen: Strange. That url of the map the doesn't have a basemap.- works fine for me.

src/components/layers/basemaps/BasemapCard.js
jenniferarnesen: Removed the spinner altogether.

src/components/layers/basemaps/BasemapCard.js
```diff
@@ -11,71 +12,65 @@ import {
...
-        >
-            <BasemapList
-                selectedID={props.basemap.id}
-                basemaps={props.basemaps}
-                selectBasemap={props.selectBasemap}
-            />
-        </LayerCard>
+        <>
+            {basemap.id === undefined ? (
+                <ComponentCover translucent>
```
turban, line 30: Remove `translucent`. The spinner looks better without it when it's no content to cover.

src/components/map/MapView.js
```diff
@@ -27,35 +28,45 @@ const MapView = props => {
...
-                    basemap={basemap}
-                    layer={splitViewLayer}
-                    controls={mapControls}
-                    feature={feature}
-                    openContextMenu={openContextMenu}
-                    resizeCount={resizeCount}
-                />
+        <>
+            {basemap.id === undefined ? (
+                <ComponentCover translucent>
```
turban, line 33: Remove `translucent`. The spinner looks better without it when it's no content to cover.

turban (commented):
Looks mostly fine now! 

Possible to keep App.js cleaner as a high level component and move the async fetchData function to a separate file?

One bug: 
1. Turn off the visibility of the basemap
2. Add a layer and save the map - basemap is set to "none"
3. Load the map using plugin.html, and the default basemap is shown (it should not show)

src/map.js
```diff
@@ -92,50 +92,70 @@ function PluginContainer() {
...
+            );
         }
     }
 
-    async function loadLayers(config) {
+    async function getBasemaps(basemapId, defaultBasemapId) {
+        try {
+            let externalBasemaps = [];
+            if (isValidUid(basemapId) || isValidUid(defaultBasemapId)) {
+                const externalLayers = await fetchExternalLayersD2();
```
turban, line 128: Possible to only fetch the external layer with the valid uid?

turban (approved):

PR #2009, opened 2022-01-18
fix: layer scope, data table filter and open as chart base url (DHIS2-12462)

turban:
Fixes: https://jira.dhis2.org/browse/DHIS2-12462

This PR fixes 3 issues only present in master: 
- Run updateLayer method in the layer scope
- Remove "removeLayer" method of org unit layer (we don't use the click event anymore)
- Set the correct base url when open as chart

### DIFF
--- a/src/util/analyticalObject.js
+++ b/src/util/analyticalObject.js
@@ -124,10 +124,14 @@ export const setCurrentAnalyticalObject = async ao => {
 };
 
 // Sets analytical object to open it in another app
-export const setAnalyticalObjectAndSwitchApp = async (layer, openAs) => {
+export const setAnalyticalObjectAndSwitchApp = async (
+    layer,
+    openAs,
+    baseUrl
+) => {
     const ao = getAnalyticalObjectFromThematicLayer(layer);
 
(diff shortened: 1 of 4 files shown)


src/util/analyticalObject.js
```diff
@@ -126,7 +126,8 @@ export const setCurrentAnalyticalObject = async ao => {
 // Sets analytical object to open it in another app
 export const setAnalyticalObjectAndSwitchApp = async (layer, openAs) => {
     const ao = getAnalyticalObjectFromThematicLayer(layer);
-    const url = `${config.appUrl}/${APP_URLS[openAs]}/#/currentAnalyticalObject`;
+
+    const url = `${process.env.DHIS2_BASE_URL}/${APP_URLS[openAs]}/#/currentAnalyticalObject`;
```
jenniferarnesen, line 130: Suggest using hook from app-runtime instead of the process.env var (in OverlayCard where this function is called):

```
import { useConfig } from '@dhis2/app-runtime'
...
const { baseUrl } = useConfig()

setAnalyticalObjectAndSwitchApp(layer, type, baseUrl)
```

turban: Fixed!

jenniferarnesen (approved):

PR #2013, opened 2022-01-20
fix: upgrade to latest @dhis2/ui version

turban:
This PR will upgrade to the latest @dhis2/ui version. It also required some minor CSS and snapshot test adjustments. 

The PR also fixes a style for the user org unit select icons: 

After this PR: 

Screenshot 2022-01-20 at 18 11 15

Before:

Screenshot 2022-01-20 at 18 11 44


jenniferarnesen:
yarn.lock contains 2 versions of the ui components: 6.20.0 and 7.14.2. The 6.20 comes from dhis2/cli-app-scripts -> dhis2/app-shell and I'm wondering if also upgrading cli-app-scripts to the latest would clean up the multiple versions. I don't think it has any impact on the app's behaviour though so not an absolute necessity.

jenniferarnesen (approved):

turban:
> yarn.lock contains 2 versions of the ui components: 6.20.0 and 7.14.2. The 6.20 comes from dhis2/cli-app-scripts -> dhis2/app-shell and I'm wondering if also upgrading cli-app-scripts to the latest would clean up the multiple versions. I don't think it has any impact on the app's behaviour though so not an absolute necessity.

turban:
Also upgraded to latest version: 
- @dhis2/analytics (21.4.2)
- @dhis2/cli-app-scripts (8.3.2)

"d2-app-scripts test" is used to run unit tests, which required som adaptions of the code.

PR #2063, opened 2022-03-01
fix: no coordinates warnings

turban:
This PR will unify no coordinate warnings across facility, thematic and earth engine layers: 

**Facility layer**: no facility coordinates found
Screenshot 2022-03-01 at 16 14 57

**Facility layer**: no catchment area coordinates found
Screenshot 2022-03-01 at 16 13 26

**Facility layer**: no facility or catchment area coordinates found
Screenshot 2022-03-01 at 16 15 12

**Org unit layer**: no org unit coordinates found
Screenshot 2022-03-01 at 16 21 26

**Org unit layer**: no catchment area coordinates found
Screenshot 2022-03-01 at 16 22 00

**Org unit layer**: no org uni or catchment area coordinates found
Screenshot 2022-03-01 at 16 21 39

**EE layer**: no org unit coordinates found
Screenshot 2022-03-01 at 16 27 57

**EE layer**: no catchment area coordinates found
Screenshot 2022-03-01 at 16 28 18

Earth Engine layer will not show two warnings at the same time - as we show one or the other.

### DIFF
--- a/src/loaders/earthEngineLoader.js
+++ b/src/loaders/earthEngineLoader.js
@@ -42,9 +42,10 @@ const earthEngineLoader = async config => {
                 if (!associatedGeometries.length) {
                     alerts.push({
                         warning: true,
-                        message: `${coordinateField.name}: ${i18n.t(
-                            'No coordinates found'
-                        )}`,
+                        message: i18n.t('{{name}}: No coordinates found', {
+                            name: coordinateField.name,
+                            nsSeparator: ';',
+                        }),
                     });
                 }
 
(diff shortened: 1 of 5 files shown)


src/loaders/earthEngineLoader.js
```diff
@@ -38,27 +39,32 @@ const earthEngineLoader = async config => {
                     })
                     .then(toGeoJson);
 
+                if (!associatedGeometries.length) {
+                    alerts.push({
+                        warning: true,
+                        message: `${coordinateField.name}: ${i18n.t(
```
jenniferarnesen, line 45: It might be possible to use a one-time ns separator so that this can be in one string (e.g. semicolon in this example):
```
message: i18n.t(`{{name}}: No coordinates found`, {name: coordinateField.name, nsSeparator: ';'})
```

turban: Changed!

PR #2078, opened 2022-03-11
fix: filter out invalid geojson

turban:
This PR will filter out GeoJSON polygons that have no coordinates `[[]]`. 

After this PR no error shows when selecting catchment areas in Kailahun:
Screenshot 2022-03-11 at 22 31 48

### DIFF
--- a/src/util/map.js
+++ b/src/util/map.js
@@ -65,7 +65,8 @@ export const toGeoJson = organisationUnits =>
         .filter(
             ({ geometry }) =>
                 Array.isArray(geometry.coordinates) &&
-                geometry.coordinates.length
+                geometry.coordinates.length &&
+                geometry.coordinates.flat().length
         );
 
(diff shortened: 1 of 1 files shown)


src/util/map.js
```diff
@@ -65,7 +65,8 @@ export const toGeoJson = organisationUnits =>
         .filter(
             ({ geometry }) =>
                 Array.isArray(geometry.coordinates) &&
-                geometry.coordinates.length
+                geometry.coordinates.length &&
```
jenniferarnesen, line 68: Are both needed or can you just have `geometry.coordinates.flat().length`

jenniferarnesen (approved):

src/util/map.js
turban: The argument for keeping both is that flat() can be an expensive operation when there are thousands of coordinates.

PR #2079, opened 2022-03-12
fix: allow popups for buildings layer while showing an error

turban:
This PR fixes two issues for the new buildings layer:
1. While we show an error from the Earth Engine API the map can still be clicked. 
2. The popup loading spinner is removed if we get an error, as data will never arrive. 

After this PR: 
buildings-error

### DIFF
--- a/src/components/map/layers/earthEngine/EarthEngineLayer.js
+++ b/src/components/map/layers/earthEngine/EarthEngineLayer.js
@@ -203,6 +203,7 @@ export default class EarthEngineLayer extends Layer {
         const { legend, aggregationType } = this.props;
         const { isLoading, popup, aggregations, error } = this.state;
 
+        /*
         if (error) {
             return (
                 <Alert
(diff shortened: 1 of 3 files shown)


src/components/map/layers/earthEngine/EarthEngineLayer.js
```diff
@@ -247,6 +245,10 @@ export default class EarthEngineLayer extends Layer {
             message = this.props.error;
         }
 
-        this.setState({ error: message, isLoading: false });
+        this.setState({
+            error: message,
+            isLoading: false,
+            aggregations: 'error',
```
edoardo, line 251: What does this do exactly?

edoardo (approved):

turban:
Setting it to something else than "loading" will remove the popup loading
spinner.

man. 14. mar. 2022, 10:43 skrev Edoardo Sabadelli ***@***.***

PR #2105, opened 2022-04-04
fix: improved map download style (DHIS2-12861, DHIS2-12862) 

turban:
Fixes: 
https://jira.dhis2.org/browse/DHIS2-12861
https://jira.dhis2.org/browse/DHIS2-12862

This PR fixes issue with the map download style: 
1. Remove padding below legend items to avoid overflow
2. Add 2 px width to map name to avoid text wrap
3. Don't hide map attribution

These style issues are related to limitations of the dom-to-image dependency: https://github.com/tsayen/dom-to-image

After this PR: 
map-d72ji

Before: 
map-xja84

### DIFF
--- a/src/components/download/DownloadDialog.js
+++ b/src/components/download/DownloadDialog.js
@@ -128,7 +128,7 @@ export class DownloadDialog extends Component {
         const skipElements = el =>
             !el.classList ||
             el.classList.contains('maplibregl-ctrl-scale') ||
-            el.classList.contains('mapboxgl-ctrl-attrib') ||
+            el.classList.contains('maplibregl-ctrl-attrib') ||
             !(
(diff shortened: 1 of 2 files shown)


src/components/download/DownloadDialog.js
```diff
@@ -149,7 +150,7 @@ export class DownloadDialog extends Component {
                 .getComputedStyle(titleEl, null)
                 .getPropertyValue('width');
 
-            titleEl.style.width = parseFloat(width) + 1 + 'px';
+            titleEl.style.width = parseFloat(width) + 2 + 'px';
```
jenniferarnesen, line 153: You could use string template here. Not crucial, but more standard these days.

turban: Fixed.

PR #2120, opened 2022-04-20
fix: d3 upgrade and custom time scale ticks

turban:
This PR will upgrade the d3-x dependencies. It required a code change to generate time ticks for the time slider. Weeks in d3 start on Sunday, while DHIS2 weeks start on Monday. The d3 `calendar` method we previously used to override this was not available after the upgrade. It is replaced by `utils/timeTicks.js` which mimics https://github.com/d3/d3-time/blob/main/src/ticks.js

With this PR the time ticks still align with the periods in the time scale: 
Screenshot 2022-04-21 at 09 33 49


src/util/timeTicks.js
```diff
@@ -0,0 +1,77 @@
+import { bisector, tickStep } from 'd3-array';
+import {
```
jenniferarnesen, line 2: Would it be possible to modify and use some of these unit tests?:  https://github.com/d3/d3-time/blob/main/test/ticks-test.js

jenniferarnesen (approved):

src/util/timeTicks.js
turban: Added a unit test to check that the time ticks start on Monday.

PR #2244, opened 2022-08-11
chore: split view as functional component

turban:
This PR refactors SplitView into a function component. 

The component still renders as before: 
Screenshot 2022-08-11 at 22 38 01

### DIFF
--- a/src/components/map/SplitView.js
+++ b/src/components/map/SplitView.js
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import PeriodName from './PeriodName';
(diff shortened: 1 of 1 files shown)


src/components/map/SplitView.js
```diff
@@ -7,104 +7,96 @@ import Layer from './layers/Layer';
...
+                    isPlugin={isPlugin}
+                    isFullscreen={showFullscreen}
+                >
+                    <Layer index={0} {...basemap} />
+                    <ThematicLayer
+                        index={1}
+                        period={period}
+                        feature={feature}
+                        {...layer}
+                        openContextMenu={openContextMenu}
```
jenniferarnesen, line 79: feature and openContextMenu are kind of "passthru props" and I'm wondering if it makes more sense to just pass them as {...rest} to ThematicLayer.

src/components/map/SplitView.js
```diff
@@ -7,104 +7,96 @@ import Layer from './layers/Layer';
...
+};
 
-        this.setState({ controls });
-    };
-
-    // From built-in fullscreen control
-    onFullScreenChange = ({ isFullscreen }) => this.setState({ isFullscreen });
-}
+SplitView.defaultProps = {
+    openContextMenu: () => {},
```
jenniferarnesen, line 99: Should this maybe be on the base Layer component instead? (also the "isRequired"  part of Line 95). I'm not an expert though on these things.

jenniferarnesen (approved):

PR #2247, opened 2022-08-15
feat: read current AO via new data store api

edoardo:
Due to `d2` being removed from DV [here](https://github.com/dhis2/data-visualizer-app/pull/2125) the old approach for user data store via `d2` had to be replaced with the new implementation provided by https://github.com/dhis2/app-service-datastore.

In order for it to work, Maps also needs to use it.

I'm not sure when this should be merged in, ideally after https://github.com/dhis2/data-visualizer-app/pull/2125 is merged, it has to be done for 2.39 anyway.


cypress/fixtures/analyticalObject.json
```diff
@@ -0,0 +1,117 @@
+{
+    "currentAnalyticalObject": {
+        "href": "http://localhost:8080/api/38/visualizations/UlfTKWZWV4u",
```
jenniferarnesen, line 3: Does it matter that the href includes localhost:8080 (for running on CI)?

edoardo: It shouldn't since this part is not used in Maps, I think the only part that is interesting for Maps is `columns`, `rows`.

turban (approved):
LGTM - we should do some extra testing when this is merged to both DV and Maps.

jenniferarnesen:
@edoardo Could you update the github workflow to use secrets for baseUrl, username and password. Like we do in our other apps. E.g.:  https://github.com/dhis2/line-listing-app/blob/dev/.github/workflows/dhis2-verify-app.yml#L115

jenniferarnesen (approved):

PR #2351, opened 2022-11-11
feat: add alert when basemap style fails to load

jenniferarnesen:
image

image

### DIFF
--- a/src/components/map/layers/BasemapLayer.js
+++ b/src/components/map/layers/BasemapLayer.js
@@ -10,13 +10,6 @@ const BasemapLayer = (
     const [layer, setLayer] = useState(null);
     const dispatch = useDispatch();
 
-    // useEffect(() => {
-    //     return () => {
-    //         console.log('cleanup', layer, context.map.getLayers().length);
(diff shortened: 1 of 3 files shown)


src/components/map/layers/BasemapLayer.js
```diff
@@ -0,0 +1,78 @@
...
+                index,
+                opacity,
+                isVisible,
+            });
+
+            setLayer(theLayer);
+            await context.map.addLayer(theLayer);
+        } catch (errorMessage) {
+            const message = `Basemap could not be added: ${errorMessage}`;
+            console.log(message);
```
turban, line 42: Remove console.log

turban (commented):
Yes, we can switch to a functional component here - it works well for basemaps (it might be harder for the other layer types without duplicating code).

I see that we get the same error message multiple times if switching basemap back and forth, or toggling the basemap visibility. Could we add a check in the ALERT_SET reducer to only add the alert if not already present?

turban (approved):
Nice and clean!

PR #2357, opened 2022-11-16
feat: add label option to include value in the label

jenniferarnesen:
New option
image

image

image
image

Value only:
image
name only:
image


src/constants/layers.js
```diff
@@ -146,6 +146,11 @@ export const getGroupSetStyleTypes = () => [
     },
 ];
 
+/* LABEL DISPLAY OPTIONS - match the definitions in maps-gl */
+export const LABEL_DISPLAY_OPTION_NAME_ONLY = 'NAME';
```
turban, line 150: Change 'NAME' to '{name}'

src/constants/layers.js
```diff
@@ -146,6 +146,11 @@ export const getGroupSetStyleTypes = () => [
...
+export const LABEL_DISPLAY_OPTION_NAME_AND_VALUE = 'NAME_AND_VALUE';
```
turban, line 151: Change 'NAME_AND_VALUE' to '{name}\n{value}'

src/constants/layers.js
```diff
@@ -146,6 +146,11 @@ export const getGroupSetStyleTypes = () => [
...
+export const LABEL_DISPLAY_OPTION_VALUE_ONLY = 'VALUE';
```
turban, line 152: Change 'VALUE' to '{value}'

src/constants/layers.js
```diff
@@ -146,6 +146,11 @@ export const getGroupSetStyleTypes = () => [
...
+
 /* LABEL STYLES */
 export const LABEL_FONT_SIZE = '11px';
```
turban, line 155: Should we change the default font size to '14px'?

turban (changes requested):
This looks mostly fine to me. Ref https://github.com/dhis2/maps-gl/pull/512#issuecomment-1337098215, could we change "_DISPLAY_OPTION_" to "_TEMPLATE_":
- LABEL_TEMPLATE_NAME_ONLY
- LABEL_TEMPLATE_NAME_AND_VALUE
- LABEL_TEMPLATE_VALUE_ONLY

turban:
One more change needed: Support for "labelTemplate" needs to be added as a property of mapView for map analytical object (type string). After that is done we should add it to "validLayerProperties" src/utils/favorites.js

turban:
I've added labelTemplate to the mapView properties that get saved. 

+ some style adjustments to align the label selections:

Screenshot 2023-01-24 at 01 17 10

turban (approved):

PR #2366, opened 2022-11-18
feat: add alert when basemap fails to load

jenniferarnesen:
image

image

Other alerts:
image

image

Delete alert added:
image

Plugin works:
image


turban (approved):
Nice refactor to use useAlert!

PR #2402, opened 2022-12-08
chore: use context to store window dimensions instead of redux

jenniferarnesen:
Part of refactor updating maps-app to same architecture as other analytics apps


turban (changes requested):

turban (approved):
Looks good!

PR #2425, opened 2023-01-10
chore: program stage select refactor

turban:
This PR will refactor the ProgramStageSelect component to use: 
- Functional component
- DHIS2 app runtime (useDataQuery)

First time I use `useDataQuery` and open for feedback!


jenniferarnesen (approved):

src/components/program/ProgramStageSelect.js
```diff
@@ -1,83 +1,85 @@
...
-            loadProgramStages,
-            onChange,
-        } = this.props
+// Load all stages for one program
+const PROGRAM_STAGES_QUERY = {
+    stages: {
+        resource: 'programs',
+        id: ({ id }) => id,
+        params: {
+            fields: ['programStages[id,displayName~rename(name)]'],
```
jenniferarnesen, line 13: does it work if it is not an array, but a string? I'm not sure what is required.

src/components/program/ProgramStageSelect.js
```diff
@@ -1,83 +1,85 @@
...
-            programStage,
-            programStages,
-            onChange,
-            className,
-            errorText,
-        } = this.props
-
-        if (!program) {
-            return null
+    // https://runtime.dhis2.nu/#/hooks/useDataQuery
```
jenniferarnesen, line 38: remove comment

src/components/program/ProgramStageSelect.js
```diff
@@ -1,83 +1,85 @@
...
+            onComplete: onProgramStagesLoad,
+            lazy: true,
         }
+    )
 
-        let items = programStages[program.id]
+    // Fetch program stages when program is changed
+    useEffect(() => {
+        refetch({ id: program.id })
+    }, [program, refetch])
```
jenniferarnesen, line 51: technically you could use `program.id` in the dependency array, but it probably doesn't matter.

src/components/program/ProgramStageSelect.js
turban: Switched to a string

src/components/program/ProgramStageSelect.js
turban: Keeping program, but added program.id to propTypes

PR #2433, opened 2023-01-11
chore: data element group select refactor

turban:
This PR will refactor the IndicatorGroupSelect component to use:
- Functional component
- DHIS2 app runtime (useDataQuery)


src/epics/dataElements.js
```diff
@@ -12,22 +10,6 @@ import * as types from '../constants/actionTypes.js'
...
-            .then((d2) =>
-                d2.models.dataElementGroups.list({
-                    fields: `id,${getDisplayPropertyUrl()}`,
-                    paging: false,
-                })
-            )
-            .then((groups) =>
-                setDataElementGroups(sortBy('name', groups.toArray()))
-            )
-            .catch(errorActionCreator(types.PROGRAMS_LOAD_ERROR))
```
jenniferarnesen, line 28: Can PROGRAMS_LOAD_ERROR also be removed?

jenniferarnesen (approved):

src/epics/dataElements.js
turban: It is still used in epics/programs.js

turban: It is still used in epics/programs.js

PR #2436, opened 2023-01-12
chore: useDataEngine instead of old apiFetch in OrgUnit components

jenniferarnesen:
* To simplify the code, OrgUnitData always does its own organisationUnitProfile query instead of getting the initial items as props. This ensures that `data` always matches the `period`.
* Only change in OrgUnitInfo is swapping out `d2` with `const { baseUrl } = useConfig()` which caused the need for curly braces and a `return`.
* Improved jest tests for OrgUnitInfo
* Deleted OrgUnitProfile and OrgUnitData jest tests and added a Cypress test instead that does some basic checks


src/components/orgunits/OrgUnitData.js
```diff
@@ -1,33 +1,43 @@
+import { useDataEngine } from '@dhis2/app-runtime'
```
turban, line 1: Could we use useDataQuery (with the the query defined above the component) instead of useDataEngine (meant for non-react contexts)?

src/components/orgunits/OrgUnitProfile.js
```diff
@@ -1,9 +1,9 @@
+import { useDataEngine } from '@dhis2/app-runtime'
```
turban, line 1: Possible to use useDataQuery as this is a react context (with a dynamic query defined above)?

turban (changes requested):

src/components/orgunits/OrgUnitProfile.js
```diff
@@ -19,24 +19,34 @@ const currentYear = String(new Date().getFullYear())
 const periods = getFixedPeriodsByType(periodType, currentYear)
 const defaultPeriod = filterFuturePeriods(periods)[0] || periods[0]
 
+const ORGUNIT_PROFILE_QUERY = {
+    profile: {
+        resource: 'organisationUnitProfile',
+        id: ({ id }) => `${id}/data`,
+        params: ({ period }) => ({
+            period,
+        }),
```
turban, line 28: We could remove period here if we do a separate data request in the OrgUnitData component. Maybe we could move all the period handling to this component?

turban (changes requested):

src/components/orgunits/OrgUnitProfile.js
jenniferarnesen: I was thinking that too.

turban (approved):
Looks good! Do you know why the cypress test is failing?

PR #2445, opened 2023-01-16
chore: functional component for FacilityDialog to avoid apiFetch

jenniferarnesen:
- switch to functional component
- break up componentDidUpdate into 2 useEffects for handling the initial defaults, and handling validation
- JSX hasn't really changed even though it looks like it. Only the format of the event handlers has changed from `setOrgUnitLevels`
to
`(val) => dispatch(setOrgUnitLevels(val))`


src/components/edit/FacilityDialog.js
```diff
@@ -39,234 +39,182 @@ import BufferRadius from './shared/BufferRadius.js'
...
         if (!id) {
             // Set default org unit level
-            if (
-                !getOrgUnitsFromRows(rows).length &&
-                facilityOrgUnitLevel &&
-                !prevState.facilityOrgUnitLevel
-            ) {
-                setOrgUnitLevels([facilityOrgUnitLevel.id])
+            if (facilityOrgUnitLevel) {
+                dispatch(setOrgUnitLevels([facilityOrgUnitLevel.id]))
```
turban, line 90: I think this should only be dispatched if there is no `getOrgUnitsFromRows(rows)`. If the user is editing am existing map, we shouldn't change the layer config.

src/components/edit/FacilityDialog.js
```diff
@@ -39,234 +39,182 @@ import BufferRadius from './shared/BufferRadius.js'
...
 
             // Set default org unit group set
-            if (
-                !organisationUnitGroupSet &&
-                facilityOrgUnitGroupSet &&
-                !prevState.facilityOrgUnitGroupSet
-            ) {
-                setOrganisationUnitGroupSet(facilityOrgUnitGroupSet)
+            if (facilityOrgUnitGroupSet) {
+                dispatch(setOrganisationUnitGroupSet(facilityOrgUnitGroupSet))
```
turban, line 95: Same as above, don't set if there already is a `organisationUnitGroupSet`.

turban (changes requested):

src/components/edit/FacilityDialog.js
jenniferarnesen: The check for id on line 87 should control for that right? An existing map will have ids for the layer and thus never run this code.  Also once you create a layer on a new map, it will also have an id, so it won't run in this case either.

 Unless you know of cases where there might not be an id

turban (approved):

PR #2454, opened 2023-01-19
chore: layer loaders refactor

turban:
Refactors the layer loaders so they are triggered by functional components allowing us to use the data engine and other hooks. 

App: Before this PR only loaded layers (with data) were included to the redux store. Now they are added to the redux store with a "isLoaded: false" config. Layer loading is triggered in `component/loaders/LayersLoader.js`.

Plugin: We follow a similar pattern, but as we don't use redux we are looping unloaded layers in `Plugin.js` using the same loader components. 

This PR is not changing the actual layer loaders code, but it adds two new possibilities: 
- We can completely rewrite the loaders (using hooks) in the _components/loaders/_ folder 
- We can keep the loaders, but pass in an engine param in the _components/loaders/_ folder

Videos and screenshots that shows that the layer still load and update: 

plugin-map

https://user-images.githubusercontent.com/548708/217273557-3e2b6773-f898-4a26-a524-10ccfa89c1c4.mov

https://user-images.githubusercontent.com/548708/217273566-e5c23545-032d-4eb7-86f6-c6508428d8eb.mov

https://user-images.githubusercontent.com/548708/217273567-e753c9a4-d167-4010-abf3-a46902010e21.mov

https://user-images.githubusercontent.com/548708/217273572-5da43f4a-473b-43ba-9a55-2a98ee1e84a2.mov

https://user-images.githubusercontent.com/548708/217273575-1067ef2d-cd7f-4920-a630-e53f4b791723.mov

https://user-images.githubusercontent.com/548708/217273578-e3c78211-a607-489e-b3ad-e8b439bb83dd.mov


src/components/loaders/LayerLoader.js
```diff
@@ -0,0 +1,36 @@
...
+    orgUnit: OrgUnitLoader,
+    thematic: ThematicLoader,
+    trackedEntity: TrackedEntityLoader,
+}
+
+const LayerLoader = ({ config, onLoad }) => {
+    const Loader = layerType[config.layer]
+
+    if (!Loader) {
+        return null
```
jenniferarnesen, line 25: should this be a warning or alert or something?

jenniferarnesen:
Looks good. It seems like actions loadLayer isn't used anymore. Can that be removed?

jenniferarnesen (approved):

src/components/loaders/LayerLoader.js
turban: Added the same console log as we had: https://github.com/dhis2/maps-app/blob/master/src/loaders/layers.js#L29

PR #2457, opened 2023-01-19
chore: load option set refactor

turban:
This PR will refactor FilterSelect and OptionSetStyle to use:
- DHIS2 app runtime (useDataQuery in a new useOptionSet hook)

Screenshot 2023-01-19 at 18 46 26
Screenshot 2023-01-19 at 18 46 00


src/components/filter/FilterSelect.js
```diff
@@ -43,14 +42,8 @@ const getOperators = (valueType, optionSet) => {
...
 
-const FilterSelect = ({
-    valueType,
-    filter,
-    optionSet,
-    optionSets,
-    onChange,
-    loadOptionSet,
-}) => {
+const FilterSelect = ({ valueType, filter, optionSet, onChange }) => {
```
jenniferarnesen, line 45: Is it possible to only pass in the `optionSetId` ?

turban:
Main changes pushed:
- FilterSelect uses `optionSetId` instead of `optionSet`
- OptionSetStyle uses `styledOptionSet` to better distinguish from the option set initially loaded
- Both components adds an extra check that option set selected is identical to the option set loaded (when the user selects another option set this is not the case for a short period)

src/components/dataItem/filter/FilterSelect.js
```diff
@@ -43,15 +42,9 @@ const getOperators = (valueType, optionSet) => {
...
-    valueType,
-    filter,
-    optionSet,
-    optionSets,
-    onChange,
-    loadOptionSet,
-}) => {
-    const operators = getOperators(valueType, optionSet)
+const FilterSelect = ({ valueType, filter, optionSetId, onChange }) => {
+    const { optionSet, fetchOptionSet } = useOptionSet()
```
jenniferarnesen, line 46: It might be possible to write it similar to the useProgramStageDataElements hook, where you just pass the id as an argument to the hook, and then you can remove the useEffect on line 59 of this file.

src/components/optionSet/OptionSetStyle.js
```diff
@@ -21,41 +21,46 @@ const addOptionStyle = (option, index) => ({
...
-}) => {
+const OptionSetStyle = ({ styledOptionSet }) => {
+    const { optionSet, fetchOptionSet } = useOptionSet()
     const [warning, setWarning] = useState()
+    const dispatch = useDispatch()
+
+    const options =
+        styledOptionSet.id === optionSet?.id ? optionSet.options : null
+
+    const styledOptions = styledOptionSet?.options
```
jenniferarnesen, line 32: The ? seems unnecessary since this object is already being accessed on line 30

src/components/dataItem/filter/FilterRow.js
```diff
@@ -18,7 +18,8 @@ const FilterRow = ({
         return null
     }
 
-    const dataItem = dimension && dataItems.filter((d) => d.id === dimension)[0]
+    const dataItem = dataItems.find((d) => d.id === dimension)
```
jenniferarnesen, line 21: What about combining these into a single line to avoid having another variable with a name similar to other variables?

src/components/optionSet/OptionSetStyle.js
```diff
@@ -21,41 +21,46 @@ const addOptionStyle = (option, index) => ({
...
 })
 
-const OptionSetStyle = ({
-    id,
-    options,
-    optionSet,
-    loadOptionSet,
-    setOptionStyle,
-}) => {
+const OptionSetStyle = ({ styledOptionSet }) => {
```
jenniferarnesen, line 24: Since redux is already being used in this component (dispatch), I'm wondering if it would make more sense to get styledOptionSet with useSelector directly instead of it being passed through the components as a prop? Really not an important thing.

src/components/dataItem/filter/FilterSelect.js
turban: Fixed!

src/components/dataItem/filter/FilterRow.js
turban: Done!

src/components/optionSet/OptionSetStyle.js
turban: Keeping the current solution as we already have this property in the parent component.

jenniferarnesen (approved):

PR #2459, opened 2023-01-25
chore: org unit tree refactor

turban:
In this PR we start to use the `OrgUnitDimension` component from `@dhis2/analytics`. 

It is wrapped in a new OrgUnitSelect component which is used by all layer types. This allowed us to greatly simplify the org unit handling in DHIS2 Maps. 

**Thematic layer**

After this PR:
Screenshot 2023-02-19 at 21 42 50

Before: 
Screenshot 2023-02-19 at 21 44 45

**Event layer**

After this PR:
Screenshot 2023-02-19 at 21 46 03

Before: 
Screenshot 2023-02-19 at 21 46 57

**Tracked entity layer**

After this PR: 
Screenshot 2023-02-19 at 21 48 49

Before: 
Screenshot 2023-02-19 at 21 49 22

**Facility layer**

After this PR:
Screenshot 2023-02-19 at 21 56 38

Before:
Screenshot 2023-02-19 at 21 56 51

**Org unit layer**

After this PR: 
Screenshot 2023-02-19 at 21 59 48

Before:
Screenshot 2023-02-19 at 22 00 24

**Earth Engine layer**

After this PR: 
Screenshot 2023-02-19 at 22 01 08

Before: 
Screenshot 2023-02-19 at 22 02 02


jenniferarnesen:
General observation: in some layers the tab says "Org Units", in others it says "Organisation Units". It should be consistent.

Also, I noticed "Use associated geometry" is a placeholder in the select, rather than a label. Was that intentional?

src/components/orgunits/OrgUnitFieldSelect.js
```diff
@@ -34,9 +34,9 @@ const OrgUnitFieldSelect = () => {
     const attribute = attributes.find((a) => a.id === orgUnitField)
```
jenniferarnesen, line 34: Maybe this file should be renamed to AssociatedGeometrySelect

src/components/orgunits/OrgUnitFieldSelect.js
```diff
@@ -34,9 +34,9 @@ const OrgUnitFieldSelect = () => {
...
 
     return (
-        <>
+        <div className={styles.orgUnitField}>
             <SelectField
-                label={i18n.t('Use associated geometry')}
+                prefix={i18n.t('Use associated geometry')}
```
jenniferarnesen, line 39: Suggest prefix:

`Associated geometry:`

src/components/orgunits/OrgUnitFieldSelect.js
```diff
@@ -34,9 +34,9 @@ const OrgUnitFieldSelect = () => {
...
                 items={[{ id: NONE, name: i18n.t('None') }, ...attributes]}
                 value={orgUnitField}
```
jenniferarnesen, line 41: Suggest having None as the default value.

src/components/orgunits/OrgUnitSelect.js
```diff
@@ -0,0 +1,127 @@
...
+                <CenteredContent>
+                    <CircularLoader />
+                </CenteredContent>
+            </div>
+        )
+    } else if (error?.message) {
+        return <Help error>{error.message}</Help>
+    }
+
+    console.log('selected', orgUnits)
```
jenniferarnesen, line 96: remove console.log

src/components/orgunits/styles/OrgUnitSelect.module.css
```diff
@@ -0,0 +1,19 @@
+.orgUnitSelect :global(.orgUnitTreeWrapper) {
+    max-height: 220px;
+}
+
+.orgUnitSelect :global(.userOrgUnitsWrapper label) {
+    margin-right: 34px;
+}
+
+.event :global(.orgUnitTreeWrapper),
```
jenniferarnesen, line 9: why do these have a larger max-height?

turban: I'm adjusting the size of the org unit tree to avoid a vertical scroll in the layer dialog that would hide some of the selects. Some layer dialogs have more selects than others as shown in the screenshots above.

src/components/orgunits/OrgUnitFieldSelect.js
turban: Fixed.

src/components/orgunits/OrgUnitFieldSelect.js
turban: Following the sketch from Joe: 

image

src/components/orgunits/OrgUnitFieldSelect.js
turban: Fixed.

src/components/orgunits/OrgUnitSelect.js
turban: Fixed.

turban:
> General observation: in some layers the tab says "Org Units", in others it says "Organisation Units". It should be consistent.

This is not changed in this PR - so I suggest that we fix it in a separate PR after asking Joe. 

> Also, I noticed "Use associated geometry" is a prefix in the select, rather than a label. Was that intentional? 

Yes, following a sketch from Joe - it saves som place and it is similar to the select above: 

Screenshot 2023-02-21 at 12 38 08

> Basically, have "None" chosen already and indicate to use that this is the default value. The addition of "(default)" was suggested by Lars for the EE import.

Changed - see above screenshot.

jenniferarnesen (approved):

PR #2370, opened 2022-11-23
chore: upgrade to app-platform and new plugin architecture

jenniferarnesen:
https://dhis2.atlassian.net/browse/DHIS2-4816

* `public` folder is removed as that doesn't fit with new architecture. The i18n_old had translations for the old Translate and Share dialogs that have been replaced and are no longer in use
* Entry point for the app is src/AppWrapper.js. This replaces app.js and Root.js
* Entry point for plugin is PluginWrapper.js. This file is identical to the same file in DV and LineList.
* Functionality for the old map.js (for the plugin) is now handled by plugin/MapContainer.js
* The old plugin/Plugin.js file has been renamed to plugin/Map.js (this code is unchanged)
* The file named plugin/Plugin.js is simply a thin wrapper that provides the D2Shim and the SystemSettingsProvider.
* Only change in AppLayout is to remove HeaderBar since that is now provided by the platform
* Removed FatalBoundaryError since that is provided by platform
* util/api.js - need to use d2 to get the baseUrl instead of `config`

How to try it out
- to run the app, use the typical steps (yarn install, yarn start)
- to try the plugin, you currently have to use it from dashboard. Follow these steps:

1. get dashboard branch `feat/map-plugin-latest`
2. in browser add the plugin override (this is for dev mode only) that looks like this:  
image
3. get this maps-app branch `chore/upgrade-to-app-platform`
4. start the dashboard on localhost:3000 `DHIS2_BASE_URL=https://debug.dhis2.org/dev yarn start`
5. start the maps-app. The app will be on localhost:3001 and the plugin on localhost:3002.  `DHIS2_BASE_URL=https://debug.dhis2.org/dev yarn start`
6. Open localhost:3000. If you use debug/dev, then you can go straight to `http://localhost:3000/#/jdCDW9aTIba` to get a dashboard with a single map on it.

PR #2461, opened 2023-01-25
chore: map plugin refactor

turban:
This PR refactors the Plugin component into a functional component. Most of the code changes are from https://github.com/dhis2/maps-app/pull/2370

When we had a class component we could call the plugin resize method using `.current.resize`. This was not possible with a functional component. As a temporal solution, the resize method is passed back to the calling component. We should find a better solution when we refactor map.js (we probably only need to support rendering a single map - not multiple as supported today). 

After this PR:
Screenshot 2023-01-25 at 21 40 40

The snapshot tests had to be updated after running `npx yarn-deduplicate && yarn`. We had two version of `@dhis2/ui`.


src/components/plugin/Plugin.js
```diff
@@ -160,15 +69,83 @@ class Plugin extends Component {
...
-                feature: null,
-            })
+                )
+            )
+
+            setContextMenu()
+        }
+    }
+
+    useEffect(() => {
```
jenniferarnesen, line 82: Can you move the `useEffect`s to the top with the other hooks?

jenniferarnesen (approved):

PR #2475, opened 2023-02-08
feat: improved map download (DHIS2-14317)

turban:
This PR improves the map download: https://dhis2.atlassian.net/browse/DHIS2-14317

What is new? 
- The download dialog is shown in the left dialog, and not in a modal above the map (`DownloadDialog`)
- Information about the map is shown in a right column next to the map  (`DownloadPanel`)
- Possible to add a map description
- Possible to add a north arrow to one of the map corners (`NorthArrowPosition`) - the arrow rotates with the map (`NorthArrow`)
- Possible to add a map legend for individual layers (`LegendLayers`)
- Possible to add an overview map (`OverviewMap`, `OverviewMapOutline`)

Bigger changes:
- A new `AppLayout` component to reduce the complexity of the `App` component
- A new 'MapPosition' component to reduce the complexity of the `MapContainer` component

After this PR: 
Screenshot 2023-02-22 at 17 48 34

With a map description:
Screenshot 2023-02-27 at 12 37 16

With multiple layers/legends and no room for overview map:
Screenshot 2023-02-27 at 12 39 05

Where individual legends are toggled and there is room for overview map: 
Screenshot 2023-02-27 at 12 40 12

Where browser window is resized to make place for overview map - the height of the overview map will adapt to the space available:
Screenshot 2023-02-27 at 12 41 12

Before: 
Screenshot 2023-02-22 at 17 49 18


src/actions/download.js
```diff
@@ -1,25 +1,13 @@
 import * as types from '../constants/actionTypes.js'
 
-// Set download state (download dialog is open)
-export const toggleDownloadDialog = (show) => ({
-    type: types.DOWNLOAD_DIALOG_TOGGLE,
-    payload: show,
+// Toggle download mode
```
jenniferarnesen, line 3: Sorry, couldn't resist :D
image

jenniferarnesen:
Regarding the "info" section. 
image

I'm just curious if the info regarding description should really be here in the download panel? It doesn't seem like the right place to be describing how to use the app.

Could you just enable the checkbox for Show Description regardless of whether a description has been set, and then show "No description".  e.g.:  
image.  The "No description" wouldn't display in the download/print.

I'm also curious what "resize your browser window to change the map dimensions" means?
@cooper-joe

jenniferarnesen:
With this selection, the download panel will still show:
image

src/actions/download.js
```diff
@@ -1,25 +1,13 @@
...
+export const toggleDownloadMode = (payload) => ({
```
jenniferarnesen, line 4: I'm not sure it makes sense to call this action a "toggle" since you are sending a value with each call. If it is toggle, then the reducer should just reverse whatever value is currently set. That said, I think toggles can be untrustworthy, so maybe just rename this to `setDownloadMode`.

src/actions/download.js
```diff
@@ -1,25 +1,13 @@
...
-    type: types.DOWNLOAD_LEGEND_SHOW_TOGGLE,
-    payload: show,
-})
-
-// Legend corner position
-export const setDownloadLegendPosition = (position) => ({
-    type: types.DOWNLOAD_LEGEND_POSITION_SET,
-    payload: position,
+// Set downlond property
+export const setDownloadProperty = (payload) => ({
```
jenniferarnesen, line 10: What about renaming this to `setDownloadConfig`? It's not clear what "property" means.

src/components/download/DownloadMode.js
```diff
@@ -0,0 +1,23 @@
...
+import styles from './styles/DownloadMode.module.css'
+
+const DownloadMode = () => {
+    const dispatch = useDispatch()
+
+    return (
+        <>
+            <div className={styles.downloadModeMenu}>
+                <Button onClick={() => dispatch(toggleDownloadMode(false))}>
+                    Exit download mode
```
jenniferarnesen, line 15: i18n

jenniferarnesen:
What about hiding the headerbar when in download mode (display: none).  That's what is done in dashboard. @cooper-joe

src/components/download/DownloadDialog.js
```diff
@@ -1,97 +1,196 @@
...
-import {
-    Modal,
-    ModalTitle,
-    ModalContent,
-    ModalActions,
-    Button,
-    ButtonStrip,
-} from '@dhis2/ui'
-import React, { useState, useMemo } from 'react'
+import { Button, ButtonStrip } from '@dhis2/ui'
```
jenniferarnesen, line 2: To me it makes more sense that this component is called "DownloadSettingsPanel"

src/components/download/DownloadPanel.js
```diff
@@ -0,0 +1,57 @@
+import cx from 'classnames'
```
jenniferarnesen, line 1: I feel like this file should be called "MapInfo" similar

jenniferarnesen:
What about including the map name as part of the  downloaded file name? When downloading the layer data (dataDownload), the downloaded file is named based on the layer name.

src/actions/download.js
turban: I agree, renamed to `setDownloadMode` and `DOWNLOAD_MODE_SET`.

src/actions/download.js
turban: Renamed to `setDowloadConfig` and `DOWNLOAD_CONFIG_SET`.

src/components/download/DownloadMode.js
turban: Good catch - i18n added

src/components/download/DownloadDialog.js
turban: Renamed to `DownloadSettings`

src/components/download/DownloadPanel.js
turban: Renamed to `DownloadMapInfo`

src/actions/download.js
turban: Comment removed :-)

turban:
> With this selection, the download panel will still show: image

Fixed.

turban:
Leaving the UX/UI questions for @cooper-joe to decide :-)

cooper-joe:
> What about hiding the headerbar when in download mode (display: none). That's what is done in dashboard. 

Yes, can mirror the functionality from Dashboards here.

> What about including the map name as part of the downloaded file name?

Agreed.

> I'm also curious what "resize your browser window to change the map dimensions" means?

I guess this is how a user gets different orientation/ratio maps, as the size of the map is relative to the browser window. I think using [css resize](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) might be a clearer way of doing this, if possible?

turban:
Fixed: 
1. Headerbar is hidden in download mode
2. Maps print mode aligns better with dashboard print mode
3. Map name (if present) is used in the download filename
4. Help for map description is shown in a tooltip

Screenshot 2023-02-27 at 14 02 40

Screenshot 2023-02-27 at 14 22 55

Screenshot 2023-02-27 at 14 23 18

src/components/core/styles/Checkbox.module.css
```diff
@@ -1,3 +1,12 @@
 .checkbox {
     margin: var(--spacers-dp12) 0 var(--spacers-dp8);
 }
+
+.checkbox label {
```
jenniferarnesen, line 5: Did you confirm that these styles don't negatively impact other places checkbox is used?

jenniferarnesen (approved):

src/components/core/styles/Checkbox.module.css
turban: Yes, I've checked all the layer dialogs and checkboxes look the same.

PR #2477, opened 2023-02-08
fix: only keep point org units when using associated geometries (DHIS2-14633)

turban:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-14633

Originally "associated geometries" where meant for catchment areas (polygons) around health facilities (point). We decided to show both geometries on the map, as it would be useful to see where the health facility is located within the catchment area. 

Showing both geometries are not ideal when: 
1. Both the main org unit geometry and the associated geometry are polygons
2. When both the main org unit geometry and the associated geometry are points

Showing both geometries (preferably with different styles) could be user option, but this requires more work and changes in the analytical object structure. 

This PR includes a fix so we only show the main org unit geometry if it is a point, and the associated geometry is not existing or not a point. For all other cases we only show the associated geometry if selected.

Associated geometries are supported in Thematic, Facility, Org unit and Earth Engine layers: 

Thematic layer show both main org unit geometry (points) and the associated geometries (polygons):

Screenshot 2023-02-09 at 11 27 03

Thematic layer showing only associated geometries (as the main geometries are also polygons):

Screenshot 2023-02-09 at 11 31 02

Org unit layer show both main org unit geometry (points) and the associated geometries (polygons):

Screenshot 2023-02-09 at 11 37 54

Org unit layer showing only associated geometries (as the main geometries are also polygons):

Screenshot 2023-02-09 at 11 39 15

Earth Engine layer show both main org unit geometry (points) and the associated geometries (polygons):

Screenshot 2023-02-09 at 11 40 13

Earth Engine layer showing only associated geometries (as the main geometries are also polygons):

Screenshot 2023-02-09 at 11 41 42

For facility layer we only show associated geometry that are polygons as the main org unit feature is already a point:

Screenshot 2023-02-09 at 11 47 23

The logic for facility layer is kept, as we only support points as the main org unit geometry. 

I plan to backport this fix 2.39 where we introduced the GeoJSON importer for associated geometries.


src/util/orgUnits.js
```diff
@@ -49,7 +49,9 @@ export const fetchOrgUnitGroupSet = (id) =>
         `/organisationUnitGroupSets/${id}?fields=organisationUnitGroups[id,name,color,symbol]`
     ).then(parseGroupSet)
 
-export const filterPointFacilities = (data) => data.filter((d) => d.ty === 1)
+export const filterPoints = (data) => data.filter((d) => d.ty === 1)
```
jenniferarnesen, line 52: `filterPoints` (and `filterPolygons`) are kind of confusing names as it's unclear if filterPoints is filtering in point, or filtering out points.  What about `getPointItems` or `getPointGeometries` or similar?

turban: Changed to `getPointItems` and `getPolygonItems`

jenniferarnesen (approved):

PR #2480, opened 2023-02-17
chore: replace epic dataDownload

jenniferarnesen:
* removed dataDownload from redux state - it's not needed.
* moved event-related functions used by both the layers/loader and dataDownload to new file `util/event.js`
* removed remaining epics as well as rxjs and redux-observable
* remove (or updated) jest tests and add cypress tests
* old d2 is still in use by the shared code. But now it is passed as a property, and we can replace it with dataQuery once we've refactored the loaders into the Loader components


turban (changes requested):
There seems to be an issue with event layer an this PR: 

Screenshot 2023-02-20 at 10 44 12

src/util/event.js
```diff
@@ -0,0 +1,126 @@
...
+    const result = await d2.models.programStage.get(layer.programStage.id, {
+        fields: `programStageDataElements[displayInReports,dataElement[id,code,${displayNameProp}~rename(name),optionSet]]`,
+        paging: false,
+    })
+
+    return result.programStageDataElements
+        .filter((el) => el.displayInReports)
+        .map((el, i) => {
+            if (i === 0) {
+                console.log('el', el)
```
turban, line 26: Remove?

turban (approved):

PR #2500, opened 2023-02-27
chore: set default org unit level and roots

turban:
This PR is partly a revert of https://github.com/dhis2/maps-app/pull/2459

In #2459 we set the default org units in the new OrgUnitSelect component. The problem with this approach is the users needs to visit the Org unit tab in the layer dialog before the default value is set. We would like users to create new layers with minimal number of clicks, and in this PR we set the default Org units when the layer dialog is opened (like previously).

A new OrgUnitsProvider component was created to only fetch the default values once. 

After this PR: 

Thematic layer with default org unit level selected: 
Screenshot 2023-02-27 at 20 43 04

Earth Engine layer with default org unit level selected: 
Screenshot 2023-02-27 at 20 44 28

Event layer with the org unit root selected:
Screenshot 2023-02-27 at 20 45 30

Tracked entity layer with the org unit root selected:
Screenshot 2023-02-27 at 20 46 20

### DIFF
--- a/src/components/Root.js
+++ b/src/components/Root.js
@@ -10,6 +10,7 @@ import { apiVersion } from '../constants/settings.js'
 import i18n from '../locales/index.js'
 import { NAMESPACE } from '../util/analyticalObject.js'
 import App from './app/App.js'
+import OrgUnitsProvider from './OrgUnitsProvider.js'
 import SystemSettingsProvider from './SystemSettingsProvider.js'
 import UserSettingsProvider, {
     UserSettingsCtx,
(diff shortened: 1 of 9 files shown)


src/components/Root.js
```diff
@@ -37,58 +38,66 @@ const d2Config = {
...
-        config={{
-            baseUrl: process.env.DHIS2_BASE_URL,
-            apiVersion,
-        }}
-    >
-        <ReduxProvider store={store}>
-            <D2Shim d2Config={d2Config} i18nRoot="./i18n_old">
-                {({ d2, d2Error }) => {
-                    if (!d2 && !d2Error) {
+const Root = ({ store }) => {
```
jenniferarnesen, line 41: Use 
const Root = ({ store }) => ( <DataProvider... )
instead of
const Root = ({ store }) => { return (<DataProvider...

src/components/edit/trackedEntity/TrackedEntityDialog.js
```diff
@@ -110,6 +116,16 @@ class TrackedEntityDialog extends Component {
                 showRelationshipsChecked: true,
             })
         }
+
+        // Set org unit tree roots as default
+        if (!rows && orgUnits.roots) {
+            if (orgUnits.roots) {
```
jenniferarnesen, line 122: Is this check necessary? It's already checked on line 121.

src/components/edit/event/EventDialog.js
```diff
@@ -131,6 +137,16 @@ class EventDialog extends Component {
             setStartDate(DEFAULT_START_DATE)
             setEndDate(DEFAULT_END_DATE)
         }
+
+        // Set org unit tree roots as default
+        if (!rows && orgUnits.roots) {
+            if (orgUnits.roots) {
```
jenniferarnesen, line 143: Is check on 143 needed since it's already checked on 142?

src/components/orgunits/OrgUnitSelect.js
```diff
@@ -3,16 +3,16 @@ import { useDataQuery } from '@dhis2/app-runtime'
...
 import { setOrgUnits } from '../../actions/layerEdit.js'
-import { DEFAULT_ORG_UNIT_LEVEL } from '../../constants/layers.js'
 import { translateOrgUnitLevels } from '../../util/orgUnits.js'
 import AssociatedGeometrySelect from './AssociatedGeometrySelect.js'
 import OrgUnitSelectMode from './OrgUnitSelectMode.js'
 import styles from './styles/OrgUnitSelect.module.css'
 
-// Fetches the root org units associated with the current user with fallback to data capture org units
+// Fetches org unit levels and the root org units associated with the current user
+// with fallback to data capture org units
```
jenniferarnesen, line 15: Is this query still needed? Looks exactly the same as the one in the Provider, so is it possible to use the `useOrgUnits` hook instead?

src/components/Root.js
turban: Fixed!

src/components/edit/trackedEntity/TrackedEntityDialog.js
turban: My bad - fixed!

src/components/edit/event/EventDialog.js
turban: My bad - fixed!

src/components/orgunits/OrgUnitSelect.js
turban: My bad - fixed!

jenniferarnesen (approved):

PR #2523, opened 2023-03-13
fix: omit fonts and images from precache

jenniferarnesen:
This will stop all the thousands of pbf files from being downloaded.
Also, other platform dependencies were upgraded to keep consistent versions.

Waiting on this PR:  https://github.com/dhis2/app-platform/pull/793


KaiVandivier (approved):
Tested on `debug/dev` and confirmed that fonts and images aren't precached any more 🙂

PR #2525, opened 2023-03-13
fix: issues found during team testing of 2.40 release

turban:
Fixes: 
1. Download mode is kept when user is downloading a map
2. Close data table if entering download mode
3. Download filename changes: Special characters are removed (not replaced with '_'. Uppercase letters and spaces are kept in the filename. We don't append a random string at the end. Filename examples with or without a map name:

Screenshot 2023-03-13 at 17 01 50

### DIFF
--- a/src/util/dataDownload.js
+++ b/src/util/dataDownload.js
@@ -45,8 +45,9 @@ export const addPropNames = (layer, data) => {
 }
 
 // Replaces anything that's not a letter, number or space
+// Multiple spaces is replaced by a single space in the last replace
 export const standardizeFilename = (name, ext) => 
-    `${name.replace(/[^a-z0-9 ]/gi, '')}.${ext}`
+    `${name.replace(/[^a-z0-9 ]/gi, '').replace(/  +/g, ' ')}.${ext}`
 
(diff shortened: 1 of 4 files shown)


src/util/dataDownload.js
```diff
@@ -44,13 +44,10 @@ export const addPropNames = (layer, data) => {
...
-// Replaces anything that's not a letter or a number, and makes it all lower case
-// If no rawName the filename will start with "map_"
-// A short random string is added to avoid multiple files with the same filename
-export const standardizeFilename = (rawName, ext) =>
-    `${
-        rawName ? rawName.replace(/[^a-z0-9]/gi, '_').toLowerCase() : 'map_'
-    }_${Math.random().toString(36).substring(7)}.${ext}`
+// Replaces anything that's not a letter, number or space
+// Multiple spaces is replaced by a single space in the last replace
+export const standardizeFilename = (name, ext) =>
```
jenniferarnesen, line 49: please add one or 2 jest tests that demonstrate what this function does.

jenniferarnesen (approved):

PR #2531, opened 2023-03-15
fix: avoid interpretation map redraws

turban:
This PR fixes issues with the map interpretations:
1. The interpretations panel (right drawer) is not opened before the map is fully loaded. This avoids redraws of the interpretation map as it uses the same layers as the main map. 
2. `InterpretationsPanel` code is simplified: the interpretation modal is open if there is an `interpretationId`. `isMapLoading` where not in use (always false), and removed.
3. Fix to support split view maps in a plugin.

After this PR:

ezgif com-video-to-gif (1)


src/components/interpretations/InterpretationsPanel.js
```diff
@@ -36,59 +27,39 @@ const InterpretationsPanel = ({
...
                     onInterpretationUpdate={() =>
                         interpretationsUnitRef.current.refresh()
                     }
                     initialFocus={initialFocus}
                     interpretationId={interpretationId}
-                    isVisualizationLoading={isMapLoading}
+                    isVisualizationLoading={false}
                     onClose={onModalClose}
-                    onResponsesReceived={() => setIsMapLoading(false)}
+                    onResponsesReceived={() => {}} // Required prop
```
jenniferarnesen, line 62: Use Function.prototype

jenniferarnesen (approved):

src/components/interpretations/InterpretationsPanel.js
turban: Fixed.

PR #2557, opened 2023-04-03
fix: new total population layer

turban:
With this PR we are using the same Earth Engine dataset for both totalt population and age/gender groups. This makes sure that the totalt population numbers are the same. 

Fixes: https://dhis2.atlassian.net/browse/DHIS2-14282

This means that we only have population values for 2020. 

As we use the same EE dataset for two layers, we couldn't rely to the `datasetId` to be unique, a new `layerId` was therefore added. 

Some code that where included to support old EE layers where removed, as these layers have not been supported for several DHIS2 versions. 

After this PR the total population in Bo is identical to selecting all age/gender groups: 

Screenshot 2023-04-03 at 12 46 52
Screenshot 2023-04-03 at 12 48 06

Before this PR the totalt population in Bo is much less: 

Screenshot 2023-04-03 at 12 50 45

We should also update the Import/Export app to reflect these changes.

### DIFF
--- a/src/util/earthEngine.js
+++ b/src/util/earthEngine.js
@@ -1,6 +1,5 @@
 import i18n from '@dhis2/d2-i18n'
 import { loadEarthEngineWorker } from '../components/map/MapApi.js'
-import { getEarthEngineLayer } from '../constants/earthEngine.js'
 import { apiFetch } from './api.js'
 import { formatStartEndDate } from './time.js'
 
--- a/src/util/earthEngine.js
+++ b/src/util/earthEngine.js
@@ -95,6 +95,11 @@ export const getPeriods = async (eeId, periodType) => {
         const name =
             periodType === 'Yearly' ? String(year) : getStartEndDate(properties)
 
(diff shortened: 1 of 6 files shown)


src/util/earthEngine.js
```diff
@@ -90,19 +89,12 @@ const getWorkerInstance = async () => {
...
-    const { periodType } = getEarthEngineLayer(eeId)
-
+export const getPeriods = async (eeId, periodType) => {
     const getPeriod = ({ id, properties }) => {
         const year = new Date(properties['system:time_start']).getFullYear()
         const name =
             periodType === 'Yearly' ? String(year) : getStartEndDate(properties)
 
-        // Remove when old population should not be supported
-        if (eeId === 'WorldPop/POP') {
```
jenniferarnesen, line 102: This eeId is still present in `constants/earthEngine.js`

turban: Reverting this removal as we have a saved map from 2016 in the demo db that still uses this dataset.

PR #2722, opened 2023-06-26
chore: dependency upgrades

turban:
This PR upgrades dependencies to the latest minor/path version.


jenniferarnesen (approved):

cypress/elements/basemap_card.js
```diff
@@ -21,6 +21,8 @@ const activeBasemap = (basemapName) => {
     cy.getByDataTest('basemaplistitem-name')
         .contains(basemapName)
         .scrollIntoView()
+
+    cy.getByDataTest('basemaplistitem-name')
```
jenniferarnesen, line 25: I think the `.contains(basemapName)` is also needed.

PR #2745, opened 2023-06-28
feat: custom calculations for thematic layer (DHIS2-15474)

turban:
This PR adds support for custom calculations for thematic layers.

Fixes: https://dhis2.atlassian.net/browse/DHIS2-15474

After this PR there is a new "Calculations" data type: 
Screenshot 2023-08-30 at 14 38 32

A new drop down allows you to select a custom calculation: 
Screenshot 2023-08-30 at 14 49 16

If there are no calculations an empty dropdown will show:
Screenshot 2023-08-30 at 15 01 06

@cooper-joe, maybe we should add a help text telling that the user can create custom calculations in the Data Visualizer app? This is not intuitive. 

Custom calculation layer added to the map: 
Screenshot 2023-08-30 at 14 55 10

Using this custom calculation: 
Screenshot 2023-08-30 at 14 46 02


cooper-joe:
> @cooper-joe, maybe we should add a help text telling that the user can create custom calculations in the Data Visualizer app? This is not intuitive.

Good idea. I suggest adding a help text under the _Calculation_ input with content `Calculations can be created in the Data Visualizer app.`

src/components/calculations/CalculationSelect.js
```diff
@@ -0,0 +1,52 @@
...
+            loading={loading}
+            items={items}
+            value={value}
+            onChange={(dataItem) => onChange(dataItem, 'calculation')}
+            className={className}
+            errorText={
+                error?.message || (!calculation && errorText ? errorText : null)
+            }
+            filterable={true}
+            dataTest="calculationselect"
```
cooper-joe, line 40: We can have a more relevant text when there are no results found for filter, `noMatchText={i18n.t('No calculations found`)}`

cooper-joe (approved):

src/components/calculations/CalculationSelect.js
turban: Changed to "No calculations found"

Screenshot 2023-08-31 at 16 07 05

turban:
Help text added: 

Screenshot 2023-08-31 at 16 06 52

jenniferarnesen (approved):

PR #2791, opened 2023-07-06
chore: update cypress to v12 and adjust config files

HendrikThePendric:
Note that this PR is not in draft because the e2e checks do not run for draft PRs. This is also why I added the "On hold" label


public/manifest.webapp
```diff
@@ -0,0 +1,22 @@
+{
```
jenniferarnesen, line 1: Is this file supposed to be added?

.github/workflows/dhis2-verify-app.yml
```diff
@@ -76,7 +76,7 @@ jobs:
     call-workflow-e2e-prod:
         if: "!contains(github.event.head_commit.message, '[skip ci]')"
         needs: [build, lint, test]
-        uses: dhis2/data-visualizer-app/.github/workflows/e2e-prod.yml@dev
+        uses: dhis2/line-listing-app/.github/workflows/e2e-prod.yml@master
         secrets:
             baseurl: ${{ secrets.CYPRESS_DHIS2_BASE_URL_DEV }}
```
jenniferarnesen, line 81: @HendrikThePendric Just a side note, I don't think `baseurl` is actually used in the shared workflow (even though it lists it as required).

i18n/cs.po
```diff
@@ -20,6 +21,10 @@ msgstr "Mapa bez názvu, {{date}}"
 
 msgid "Map \"{{- name}}\" is saved."
 msgstr "Mapa \"{{- name}}\" je uložena."
+msgstr ""
```
jenniferarnesen, line 24: This looks like one too many `msgstr` for this string, right?

.github/workflows/dhis2-verify-app.yml
HendrikThePendric: @jenniferarnesen I think the action that computes the versions uses it right?

i18n/cs.po
jenniferarnesen: This might cause the file to be corrupt. Suggest rebasing this branch and dropping the commit with the translations

.github/workflows/dhis2-verify-app.yml
jenniferarnesen: I don't think so. If you search for `baseurl`, it's only occurrence is in the list of required vars: https://github.com/dhis2/line-listing-app/blob/master/.github/workflows/e2e-prod.yml

i18n/cs.po
HendrikThePendric: Well spotted, thanks. I mustn't have resolved the conflicts correctly during my previous rebase. Rebasing again and dropping this commit seems to have worked fine. I'm not completely sure how it ended up in there in the first place.

.github/workflows/dhis2-verify-app.yml
HendrikThePendric: I thought it was needed by the `compute-dev-version` job in `e2e-dev.yml`:
https://github.com/dhis2/line-listing-app/blob/db3d74809be0677c05a85a5a2a47947616fdc22c/.github/workflows/e2e-dev.yml#L24-L36

And that the shared action which is being used requires it:
https://github.com/dhis2/action-instance-version/blob/main/action.yml#L4-L6

HendrikThePendric: Oh hold up. I see what you mean now.... `e2e-prod` does *not* need it. `e2e-dev` does, but we're not talking about that file. Sorry.

HendrikThePendric: I'll just keep it there for now. I dream of a time when we have a bunch of shared workflows for all our apps. Hope it will be this autumn.

PR #2796, opened 2023-07-07
fix: remove internal alert stack and use the platform AlertStack

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-15438

The alerts were duplicated because there was both an internal alert stack/bar system as well as the built in one from app-runtime. Solution was to remove the internal alert stack/bar. This meant that the alerts coming from the layer loaders had to be refactored so that loader alerts are no longer are added to redux. Instead the component LayerLoaders use the app-runtime hook (useAlert).

Other changes:
* `skipPluginLogic: true` has been added to d2.config.js because of the cli-app-scripts upgrade. Previously we had set a resolution for the dependency to workaround this bug: https://dhis2.atlassian.net/browse/DHIS2-16916. Now, we can remove the resolution as long as we include this config.

- [ ] Testing: confirm that alert messages still show in the dashboard item legend.

Before (see video in Jira issue)

After: 
https://github.com/dhis2/maps-app/assets/6113918/309a29c4-e467-47a6-88e3-0e63ef5534b4

For reference, these are all the layer messages (where majority of changes are):

Earth Engine alerts:
image
image
image

Event Layer alerts:
image
image
image
image

Facility layer alerts
image
image

OrgUnit Layer alerts
image
image
image

Thematic Layer alerts
image
image
image
image

TE Layer alerts
image
image


src/loaders/facilityLoader.js
```diff
@@ -80,11 +82,8 @@ const facilityLoader = async (config) => {
 
         if (!associatedGeometries.length) {
```
BRaimbault, line 83: Similar to EarthEngineLoader.js L65, we get alerts for both OU missing geom and missing associated geom.
image

src/loaders/orgUnitLoader.js
```diff
@@ -37,11 +42,8 @@ const orgUnitLoader = async (config) => {
             .catch((error) => {
                 if (error && error.message) {
```
BRaimbault, line 43: API related errors do not seem to have a message property.

src/loaders/orgUnitLoader.js
```diff
@@ -37,11 +42,8 @@ const orgUnitLoader = async (config) => {
...
-                        critical: true,
-                        message: i18n.t('Error: {{message}}', {
-                            message: error.message,
-                            nsSeparator: ';',
-                        }),
+                        code: ERROR_CRITICAL,
+                        message: error.message,
                     })
                 }
             }),
```
BRaimbault, line 49: We should also catch errors from `getOrgUnitLevels(d2)` next line.

src/components/loaders/useLoaderAlerts.js
```diff
@@ -0,0 +1,87 @@
...
+        onHidden: loaderAlertAction,
+    })
+    const infoAlert = useAlert(ALERT_MESSAGE_DYNAMIC, {
+        info: true,
+        duration: 5000,
+        onHidden: loaderAlertAction,
+    })
+    const noDataAlert = useAlert(ALERT_MESSAGE_DYNAMIC, {
+        info: true,
+        duration: 5000,
```
BRaimbault, line 28: It seems a bit inconsistant to me that no geometry is getting the warning treatment and is persisting while no data is only an info and disappears.

d2.config.js
```diff
@@ -20,6 +20,8 @@ const config = {
     },
 
     coreApp: true,
+
+    skipPluginLogic: true,
```
BRaimbault, line 24: Why is this necessary?

src/loaders/trackedEntityLoader.js
BRaimbault, line 1: There are a number of other errors we could catch, but that's probably beyond the scope of this PR.

src/loaders/thematicLoader.js
```diff
@@ -64,10 +70,8 @@ const thematicLoader = async (config) => {
                 ? {
                       alerts: [
                           {
-                              critical: true,
-                              message: `${i18n.t('Error')}: ${
-                                  error.message || error
-                              }`,
+                              code: ERROR_CRITICAL,
+                              message: error.message || error,
```
BRaimbault, line 74: This works well, maybe it should be used in other layer loaders.

src/loaders/thematicLoader.js
```diff
@@ -180,26 +184,21 @@ const thematicLoader = async (config) => {
...
             alerts.push({
-                warning: true,
-                message: `${name}: ${i18n.t('No data found')}`,
+                code: INFO_NO_DATA,
+                message: name,
             })
         }
     }
 
     if (coordinateField && !associatedGeometries.length) {
```
BRaimbault, line 198: Similar to EarthEngineLoader.js L65, we get alerts for both OU missing geom and missing associated geom.

image

BRaimbault (changes requested):
I belive some of the comments need to be addressed. Keep me posted.

d2.config.js
jenniferarnesen: Good catch, info added to the PR description.

src/components/loaders/useLoaderAlerts.js
jenniferarnesen: I can definitely reconsider that. Currently "no data" is a warning, but that is not in line with our other apps, which simply present the empty chart/table/linelist, and "no data" not necessarily meaning there is something wrong. That's why I changed it. One improvement could be to add a "no data" indicator to the Layer Card? That could be implemented later once Joe is back.

You raise a good question as to whether "no geometry" and "no data" should behave the same. To me, they are slightly different. I was thinking of geometry as more of an instance configuration, the user was clearly expecting those to be defined and they weren't.

I could switch "no data" back to a warning, to be consistent with current behavior. Then we could address the way we do alerts altogether, once Joe is back.

src/loaders/earthEngineLoader.js
```diff
@@ -62,22 +64,14 @@ const earthEngineLoader = async (config) => {
 
             if (!features.length) {
```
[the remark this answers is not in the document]
jenniferarnesen: I was also not crazy about the double warning, however I was trying not to change the actual logic that generates the warnings.

src/loaders/trackedEntityLoader.js
jenniferarnesen: Yes. Let's address this when we can refactor the way we present errors to the user.

src/loaders/earthEngineLoader.js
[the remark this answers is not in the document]
BRaimbault: Also for consistency, maybe the OU message could read: 
"Selected org units: No coordinates found"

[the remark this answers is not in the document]
jenniferarnesen: Actually, one reason to leave this as is for now is because all the translations, would be made obsolete.

src/loaders/facilityLoader.js
jenniferarnesen: This also was on purpose - to avoid logic changes.

src/loaders/orgUnitLoader.js
jenniferarnesen: Let's do this after we have removed d2 (part of the remove schemas effort)

src/loaders/orgUnitLoader.js
jenniferarnesen: Fixed

PR #2934, opened 2023-08-14
feat: read earth engine layers from data store

turban:
Depends on https://github.com/dhis2/maps-gl/pull/548


turban:
Replaced by https://github.com/dhis2/maps-app/pull/2992

PR #2937, opened 2023-08-18
fix: accept both lowercase and camelCase interpretationId

jenniferarnesen:
Needed for: https://dhis2.atlassian.net/browse/DHIS2-15441

The analytics interpretation component sends a camelcase `interpretationId` url parameter. This same format is used in DV and LL.

Including both lowercase and camelcase ensures that this version of maps isn't dependent on also having the updated dashboard app


turban (commented):
I suggest that we just change it to 

`const urlInterpretationId = getUrlParameter('interpretationId')`

We don't set this URL parameter in the maps app.

jenniferarnesen:
> I suggest that we just change it to
> 
> `const urlInterpretationId = getUrlParameter('interpretationId')`
> 
> We don't set this URL parameter in the maps app.

I would prefer to handle both cases for now because  the `view interpretation` link in the dashboard will be broken if we remove handling of the lowercase `interpretationid`.

As soon as we release the dashboard with the updated interpretations component from analytics, then we can remove the lowercase `intepretationid` here.

turban (approved):

PR #2940, opened 2023-08-28
chore: useCachedQueryProvider to ensure required data is loaded for rendering

jenniferarnesen:
By using the CachedQueryProvider we get the following benefits:
1. Remove SystemSettingsProvider and UserSettingsProvider
2. `loading` status is handled so maps components don't need to check if systemSettings or userSettings are loaded
3. Stores the basemaps and layerTypes in its state so they can be removed from maps redux (these lists don't change during the user session - so redux is overkill)

Notes:
* Cypress tests have been added that check some of the systemSettings and userSettings
* The only change in most of the 'Select' components is that the `nameProperty` comes from the CachedQueryProvider instead of the User/SystemSettingsProvider
* AppWrapper and App are simplified since 'loading' doesn't need to be checked

This work is being done now to simplify implementing the routing


src/components/layers/basemaps/BasemapList.js
```diff
@@ -1,26 +1,29 @@
...
-                />
-            ))}
-    </div>
-)
+const BasemapList = ({ selectedID, selectBasemap }) => {
+    const { basemaps } = useCachedDataQuery()
+    return (
+        <div className={styles.basemapList} data-test="basemaplist">
+            {basemaps
+                .filter((basemap) => layerTypes.includes(basemap.config.type))
```
jenniferarnesen, line 13: @turban Is there a reason why we shouldn't just do this filtering at the very beginning when the basemap list is first created? here: https://github.com/dhis2/maps-app/pull/2940/files#diff-4c1bfe25cf9d54e82006ccbc93b64f4b14e5e60dc2139eaf3c9c438671498117R30

As in, are there cases where we need the basemaps that dont have a `type`? Maybe for backwards compatibility where old maps are using these basemaps...

PR #2951, opened 2023-09-06
fix: post a data statistics view when map is open or saved

jenniferarnesen:
Fixes one part of https://dhis2.atlassian.net/browse/DHIS2-15800

When a map is opened, saved or save as, then POST a VISUALIZATION_VIEW event. The result is that the count in the Side panel "About this map" will show the correct number of views for the map.

image

### DIFF
--- a/src/actions/map.js
+++ b/src/actions/map.js
@@ -55,7 +55,7 @@ export const tOpenMap =
             // record visualization view
             dataEngine.mutate(dataStatisticsMutation, {
                 variables: { id: mapId },
-                onError: (error) => console.log('Error: ', error),
+                onError: (error) => console.error('Error: ', error),
             })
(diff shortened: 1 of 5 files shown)


src/actions/map.js
```diff
@@ -51,6 +52,12 @@ export const tOpenMap =
         try {
             const map = await fetchMap(mapId, dataEngine, keyDefaultBaseMap)
 
+            // record visualization view
+            dataEngine.mutate(dataStatisticsMutation, {
+                variables: { id: mapId },
+                onError: (error) => console.log('Error: ', error),
```
turban, line 58: Use same console.error as for useDataMutation(dataStatisticsMutation, ...) in FileMenu,js?

PR #2961, opened 2023-09-13
fix: force table cell content to wrap to prevent overly wide modal

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-5920

* Adding a `max-width` to the popup results in table cell content wrapping when needed.
* Event popup styles migrated from maps-gl. Styles are still in maps-gl as well so as not to break 2.38 and 2.39.
* Removed unused styles in popup.css

Before:
image

After:
image


turban (approved):
Approved with a comment: 
Keep th header cells and maybe add scope="row"? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th

Good for screen readers, but I'm not sure if they're able to access this popup anyway.

PR #2977, opened 2023-09-16
chore: period type select refactor

turban:
This PR will refactor the period type selector: 
- `clearPeriod` boolean is changed to `keepPeriod` boolean to simplify the code. If `keepPeriod` is true, we will not clear the period when the period type is changed.
- `hiddenPeriods` systems settings is read in the `PeriodTypeSelect` component so we don't need to pass it in as a property.
- It is possible to show the period type select without the "Relative" option. We will use this later for some Earth Engine layers not supporting relative periods. 
- The code setting the default period type was simplified: By default it will be set to the first option (which will be still be "Relative" for thematic layers). If we load a saved map, we will set the period type to "Relative" (first option) if a relative period is used. 

For thematic layers there should be no change in behaviour with this code change.


src/actions/layerEdit.js
```diff
@@ -179,10 +179,10 @@ export const setPeriodName = (periodName) => ({
 })
 
 // Set period type (thematic)
-export const setPeriodType = (periodType, clearPeriod = true) => ({
+export const setPeriodType = (periodType, keepPeriod) => ({
```
jenniferarnesen, line 182: This function is used in TrackedEntities PeriodTypeSelect, but without the second argument. Will this change the behavior for the TE Dialog since the default will be reversed?

turban: It should not be affected: if we the second argument is not passed we will clear the period, and this is happening in `src/reducers/layerEdit.js`.

jenniferarnesen (approved):

PR #2982, opened 2023-09-25
chore: color scale refactor

turban:
For map analytical objects we store color scales as strings with a comma delimiter. We often need to split the color string into an array to operate on them, which leads to code that can be hard to read. In this PR we keep color scales and palettes as arrays throughout the codebase. As the analytical objects only support color scales as string, we still need to convert them when saving/loading a map.


src/util/getMigratedMapConfig.js
```diff
@@ -79,15 +79,27 @@ const upgradeGisAppLayers = (config) => {
...
+            (view) =>
+                view.layer === 'boundary' || typeof view.colorScale === 'string'
+        )
+    ) {
+        return {
+            ...config,
+            mapViews: config.mapViews.map((view) => ({
+                ...view,
+                layer: view.layer === 'boundary' ? 'orgUnit' : view.layer,
+                colorScale:
```
jenniferarnesen, line 96: Could you add a jest test for this?

turban: Jest test added for colorScale translation, we already had a test for changing layer name.

src/constants/earthEngine.js
```diff
@@ -254,7 +268,14 @@ export const earthEngineLayers = () => [
...
             max: 1500,
-            palette: '#ffffd4,#fee391,#fec44f,#fe9929,#d95f0e,#993404', // YlOrBr
+            palette: [
+                '#ffffd4',
+                '#fee391',
+                '#fec44f',
+                '#fe9929',
+                '#d95f0e',
+                '#993404',
+            ], // YlOrBr
```
jenniferarnesen, line 278: Is there a color that neatly describes this palette? (like Reds, Blues as the others have)

PR #2987, opened 2023-09-26
chore: update the toolbar feature commit with a fix for interpretationId url parameter

jenniferarnesen:
With the changes from the toolbar update, the Interpretations panel might not be loaded by default. So when the url included the interpretationId, no component was listening for it, resulting in the interpretation modal not being opened.  The url param listener is thus moved to App. Tests added

### DIFF
--- a/src/components/app/DetailsPanel.js
+++ b/src/components/app/DetailsPanel.js
@@ -12,19 +12,11 @@ const DetailsPanel = ({ interpretationsRenderCount }) => {
     const interpretationId = useSelector((state) => state.interpretation?.id)
 
     const getContent = () => {
-        if (interpretationId) {
+        if (interpretationId || (detailsPanelOpen && !viewOrgUnitProfile)) {
             return <Interpretations renderCount={interpretationsRenderCount} />
         }
 
-        if (!detailsPanelOpen) {
-            return null
-        }
-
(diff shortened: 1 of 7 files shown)


src/components/app/DetailsPanel.js
```diff
@@ -9,8 +9,13 @@ import styles from './styles/DetailsPanel.module.css'
 const DetailsPanel = ({ interpretationsRenderCount }) => {
     const detailsPanelOpen = useSelector((state) => state.ui.rightPanelOpen)
     const viewOrgUnitProfile = useSelector((state) => state.orgUnitProfile)
+    const interpretationId = useSelector((state) => state.interpretation.id)
```
turban, line 12: Add question mark `state.interpretation?.id`

src/components/app/__tests__/Detailspanel.spec.js
```diff
@@ -0,0 +1,153 @@
...
+/*
+Yes, Yes, Yes
+Yes, Yes, No
+Yes, No, Yes
+Yes, No, No
+No, Yes, Yes
+No, Yes, No
+No, No, Yes
+No, No, No
+*/
```
turban, line 34: Remove

turban (approved):

PR #3034, opened 2023-11-02
fix: release candidate 100.2.1

jenniferarnesen:
Bug fixes:
* https://dhis2.atlassian.net/browse/DHIS2-15953 External basemap not shown on dashboard
* https://dhis2.atlassian.net/browse/DHIS2-16024 Cannot sort layers in map
* https://dhis2.atlassian.net/browse/DHIS2-16078 Adding external layer results in endless spinner

Chores:
Refactored settings providers to instead use the CachedQueryProvider. This also removes basemaps and layer types (e.g. thematic...) from redux. The CachedQueryProvider fetches and holds the lists of basemaps and layer types. This refactor touched a lot of files, mostly because selector components now get the user and system settings from the new provider. So good idea to test these settings.


src/components/plugin/Plugin.js
```diff
@@ -23,6 +26,28 @@ const d2Config = {
...
+}
+
+const providerDataTransformation = ({ systemSettings }) => {
+    return {
+        systemSettings: Object.assign(
+            {},
+            DEFAULT_SYSTEM_SETTINGS,
+            systemSettings,
+            {
+                hiddenPeriods: getHiddenPeriods(systemSettings),
```
turban, line 45: Is `hiddenPeriods` needed in the plugin?

turban (approved):

src/components/plugin/Plugin.js
jenniferarnesen: Good point. I don't think so. Also, I'm going to use useCallback here to get rid of the re-rendering in the plugin.

PR #3072, opened 2023-12-04
feat: disable download button until map is rendered

turban:


src/components/download/DownloadSettings.js
```diff
@@ -205,7 +236,11 @@ const DownloadSettings = () => {
                                     : i18n.t('Close')}
                             </Button>
                             {isSupported && (
-                                <Button primary onClick={onDownload}>
+                                <Button
+                                    primary
+                                    disabled={!isRendered}
```
HendrikThePendric, line 241: Great! In puppeteer I can wait until the button is enabled and then take a screenshot.

HendrikThePendric (approved):
I'll refrain from commenting on any other part of this PR than the disabled attribute of the button, which is exactly as we agreed, so approved from my side.

turban:
Added an `isPushAnalytics` param to the url. if this is not set to `true`, the `Download` button will not be disabled while the maps is loading/rendering.

src/util/history.js
```diff
@@ -34,6 +34,11 @@ const getHashUrlParams = (loc) => {
     return params
 }
 
+const getHashUrlParam = (key) => {
+    const params = getHashUrlParams(history.location)
+    return params[key]
+}
+
```
HendrikThePendric, line 41: I added this convenience method for retrieving the value of a single query parameter by name. Note that since the `parseBooleans` option is enabled in `getHashUrlParams` (note the plural `s`), we get actual booleans for things like `getHashUrlParam('isPushAnalytics')` when the query param looks like this `?isPushAnalytics=true`.

src/components/layers/LayersToggle.js
```diff
@@ -31,7 +31,7 @@ LayersToggle.propTypes = {
 export default connect(
     (state) => ({
         isOpen: state.ui.layersPanelOpen,
-        isDownload: state.download.downloadMode,
+        isDownload: !!state.download.downloadMode,
```
HendrikThePendric, line 34: Without this `isDownload` can sometimes be `undefined`, like when clicking the "Exit download" button. This caused a prop-types error. With this `!!` `isDownload` is always a boolean and the error is not triggered. Arguably, the fix would actually lie in the reducer that produces this state, but I just went for this simple solution.

jenniferarnesen (approved):

PR #3075, opened 2023-12-08
chore: use more verbose style in LayersLoader to clarify logic

jenniferarnesen:
Because of the double-negatives and multiple clauses in this statement, I found the code hard to understand. So I re-wrote it in a more verbose style.  The logic is the same: you can run the tests on the old code and it will pass.


turban (approved):
Nice!

PR #3076, opened 2023-12-09
fix: load extended data whenever an event layer is shown in datatable

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-16269

The reason this was happening was because the layer Id was not being checked before showing the new data table, so the effect to load extended was not triggered.

Instead of having data table trigger loading the extended data through an useEffect, load the extended data in the Event Loader, and have datatable wait until that data is loaded.

Main changes are in:
1. LayersLoader (checking if extended data needs loading): https://github.com/dhis2/maps-app/pull/3076/files#diff-44bc0dab44d3316d2cdc1282fbbc52adeb00bfa4422fa9b5173afa6971ca4227R28
2. DataTable (removing the loading effect)


jenniferarnesen:
Fixed for v40 in https://github.com/dhis2/maps-app/pull/3015

PR #3079, opened 2023-12-12
feat: new precipitation and temperature layers + refactor

turban:


src/constants/earthEngineLayers/buildings_GOOGLE.js
```diff
@@ -0,0 +1,30 @@
+import i18n from '@dhis2/d2-i18n'
+import { EARTH_ENGINE_LAYER } from '../layers.js'
+
+export default {
+    layer: EARTH_ENGINE_LAYER,
+    format: 'FeatureCollection',
+    layerId: 'GOOGLE/Research/open-buildings/v1/polygons',
+    img: 'images/buildings.png',
+    datasetId: 'GOOGLE/Research/open-buildings/v1/polygons',
+    name: i18n.t('Building footprints'),
```
jenniferarnesen, line 10: Are you sure these i18n.t functions are valid JSON?

jenniferarnesen (changes requested):

src/constants/earthEngineLayers/buildings_GOOGLE.js
turban: Not valid JSON - we need to find another translation solution if we start to save these configs in a data store.

src/components/edit/earthEngine/StyleSelect.js
```diff
@@ -48,22 +48,19 @@ const StyleSelect = ({ unit, params, setParams }) => {
     return (
         <div>
             <p>
-                {i18n.t('Unit: {{ unit }}', {
-                    unit,
-                    nsSeparator: '|', // https://github.com/i18next/i18next/issues/361
-                })}
+                {i18n.t('Unit')}: {unit}
```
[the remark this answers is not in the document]
turban: Not when unit variable is kept outside the i18n.t

src/components/edit/earthEngine/PeriodSelect.js
```diff
@@ -2,20 +2,24 @@ import i18n from '@dhis2/d2-i18n'
...
     onChange,
+    onError,
     errorText,
     className,
 }) => {
+    const [periods, setPeriods] = useState()
+    const [yearPeriods, setYearPeriods] = useState()
     const [year, setYear] = useState()
-    const byYear = periodType === 'Custom'
+    const byYear = periodType === 'BY_YEAR' || periodType === 'EE_MONTHLY'
```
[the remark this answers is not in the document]
turban: Fixed

src/util/earthEngine.js
```diff
@@ -15,34 +25,46 @@ export const getStartEndDate = (data) =>
...
+
+const nonDigits = /^\D+/g
+
+// Only used for backward compatibility
+export const getPeriodFromFilter = (filter, datasetId) => {
     if (!Array.isArray(filter) || !filter.length) {
         return null
     }
 
+    const isNightTimeLights = datasetId === 'NOAA/DMSP-OLS/NIGHTTIME_LIGHTS'
```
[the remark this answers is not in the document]
turban: Fixed

src/util/favorites.js
```diff
@@ -150,13 +150,14 @@ const models2objects = (layer) => {
         )
         layer.config = JSON.stringify(eeConfig)
 
+        config.config = JSON.stringify(eeConfig)
```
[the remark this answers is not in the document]
turban: Fixed

src/constants/earthEngineLayers/buildings_GOOGLE.js
jenniferarnesen: You set this as resolved, but the json is still invalid, right? What about using constants, and then using a function to return the i18n string at runtime?

`name: BUILDING_FOOTPRINTS,`

```
const getEELayerName = name => {
    const nameMap ={
        BUILDING_FOOTPRINTS: i18n.t('Building footprints'),
        ...
    }

    return nameMap[name]
}
```

src/util/earthEngine.js
```diff
@@ -2,11 +2,23 @@ import i18n from '@dhis2/d2-i18n'
...
 import { formatStartEndDate } from './time.js'
+import { EE_MONTHLY } from '../constants/periods.js'
+import { legacyNighttimeDatasetId } from '../constants/earthEngineLayers/legacy/nighttime_DMSP-OLS.js'
 
 export const classAggregation = ['percentage', 'hectares', 'acres']
 
 export const hasClasses = (type) => classAggregation.includes(type)
 
+// Translates from dynamic to static filters
+export const translateFilters = (filters, ...args) =>
```
jenniferarnesen, line 13: A function like this should really have some unit tests. Try asking gh copilot to write some! You might have to make some adjustments, but it usually provides a really good start.

Also, what about just naming the function according to what it does, e.g.
getStaticFiltersFromDynamic
convertFiltersFromDynamicToStatic

src/util/favorites.js
```diff
@@ -133,15 +133,15 @@ const models2objects = (layer) => {
         layer.rows = layer.rows.map(cleanDimension)
     }
 
-    if (layerType === EARTH_ENGINE_LAYER) {
-        const { layerId: id, band, params, aggregationType, filter } = layer
+    if (layer === EARTH_ENGINE_LAYER) {
```
jenniferarnesen, line 136: I think you want to check `layerType` instead of `layer` (see that layer.layer was renamed on line 124)

src/constants/earthEngineLayers/buildings_GOOGLE.js
turban: I don't think this needs to be valid JSON as layer configs are still included in the app.

src/util/earthEngine.js
turban: Will fix!

turban: Changed to more descriptive names:
- getStaticFilterFromPeriod
- getStaticFiltersFromDynamic

Added some unit tests for the `getStaticFilterFromPeriod` function (which also includes `getStaticFiltersFromDynamic`)

src/util/favorites.js
turban: Fixed

jenniferarnesen (approved):

PR #3088, opened 2023-12-20
feat: add user-created custom calculations to maps

jenniferarnesen:
Implements https://dhis2.atlassian.net/browse/DHIS2-15474

Select "calculations" as the data type:
image

User has access only to their own calculations

If user deletes a calculation, and it was used in a map, the maps app displays an error:
image

Form validation (tried clicking "Add Layer" without choosing a calculation)
image


src/constants/dimension.js
```diff
@@ -64,6 +64,12 @@ export const dimConf = {
         objectName: 'pi',
         itemType: 'PROGRAM_INDICATOR',
     },
+    calculation: {
+        value: 'expressionDimensionItems',
+        dimensionName: 'dx',
+        objectName: 'ed', // Created by Bjorn, don't seem to be in use when the map is saved
```
jenniferarnesen, line 70: @turban could you expand on your comment a bit (or explain it in the response here in gh?

jenniferarnesen:
> https://pr-3088.maps.netlify.dhis2.org

https://pr-3088--dhis2-maps.netlify.app/

PR #3125, opened 2024-02-14
fix: navigating by changing url directly ignored download mode and interpretation id param

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-16810

Always dispatch `setDownloadMode` and `setInterpretationId` based on the current location parameters.


HendrikThePendric (approved):
I have verified this to be working as expected, the changes are looking good, and there are even tests in place 🚀

PR #3143, opened 2024-03-07
fix: plugin was loading geojson layers multiple times

jenniferarnesen:
Previously, the LayerLoader was getting rerendered many times, because until all the layers were marked as loaded, the check on line 91 (old code) was true. This might not have been much of a problem before, but with the geojson layers having to fetch the data, there is a delay before all the layer data is received, so then the component gets rerendered in that time, firing another request for the geojson.

The bug I was seeing is that if there are multiple geojson layers on the map, the layers would flash each time the request was made. Potentially 20-30 times. In the worst case, it resulted in a javascript execution exception

https://github.com/dhis2/maps-app/assets/6113918/b870c80b-1236-488c-a73f-e1b0c8a9e8ee

The change is to instead keep track of the loaded layers in a ref instead of state, which prevents the component from rerendering the LayerLoader when some layers are still loading. The component only needs to render twice: once when the layers are not loaded, and again once they are all loaded. This is kept track with a new state variable `mapIsLoaded`

after:

https://github.com/dhis2/maps-app/assets/6113918/badfeb82-d157-428a-a2d1-95b46993e6ee


turban (approved):
LGTM. For later, maybe these loaders should be hooks to make the code cleaner?

PR #3145, opened 2024-03-07
fix: show no data message when geojson feature has no data

jenniferarnesen:
Geojson layers may or may not have data associated with them. Show an appropriate messsage if there is no data.

Screenshot shows both the empty data table at the bottom, which would have shown all the rows of data for the layer. The feature panel at the right would have shown the data only for the feature that was clicked on. This is why the messages are slightly different.

image

### DIFF
--- a/src/components/feature/FeatureProfile.js
+++ b/src/components/feature/FeatureProfile.js
@@ -16,7 +16,7 @@ const FeatureProfile = () => {
         if (Object.keys(data).length === 0) {
             return (
                 <div className={styles.noData}>
-                    {i18n.t('No data to show for this feature')}
+                    {i18n.t('No data to show for this feature.')}
                 </div>
             )
         }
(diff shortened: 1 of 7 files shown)


src/components/feature/FeatureProfile.js
```diff
@@ -12,6 +12,33 @@ const FeatureProfile = () => {
...
+                                    <td>{data[key]}</td>
+                                </tr>
+                            ))}
+                        </tbody>
+                    </table>
+                </div>
+            </div>
+        )
+    }
+
```
HendrikThePendric, line 41: Extracting to its own `DrawerContent` component would be a bit cleaner and in this case it'd be quite easy to do because the component would only need the `data` prop.
But this is not critical at all.

HendrikThePendric (approved):
Left a small suggestion

PR #3149, opened 2024-03-08
fix: better error responses and cypress tests

jenniferarnesen:


src/components/layers/toolbar/LayerToolbar.js
```diff
@@ -35,10 +35,10 @@ const LayerToolbar = ({
             <IconButton
                 tooltip={!hasError ? i18n.t('Toggle visibility') : null}
                 onClick={toggleLayerVisibility}
-                className={{
+                className={cx(styles.visibilityButton, {
```
edoardo, line 38: Is this style defined already or is it new?

edoardo (approved):

src/components/layers/toolbar/LayerToolbar.js
jenniferarnesen: Right, `visibilityButton` isn't needed.

PR #3151, opened 2024-03-12
fix: various fixes after release testing

jenniferarnesen:
1. Limit stroke width to 0-10
2. Tab length should only take the space of the tab title
image

3. Switching from invalid to valid layer when data table is open should update the table
https://github.com/dhis2/maps-app/assets/6113918/1389f0c7-b8f8-43a3-ad31-ad6422d86fe5

4. Highlight features when data table has a filter

https://github.com/dhis2/maps-app/assets/6113918/7b35606d-c727-43d4-b0e6-592442e5bd37

5. Improve legend for geojson layers
image

6. Map should show correct point radius for points
25px (green point):
image
5px (green point):
image

7. Set max line weight in legend to 5
Before:
image

After:
image

More understandable error messages. Discussed with Joe and we agreed to always output the same simple message. The administrator can then investigate by checking the network tab and learning more about the actual error.
image

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2024-03-13T09:09:37.325Z\n"
-"PO-Revision-Date: 2024-03-13T09:09:37.325Z\n"
+"POT-Creation-Date: 2024-03-13T10:31:57.389Z\n"
+"PO-Revision-Date: 2024-03-13T10:31:57.389Z\n"
 
 msgid "Untitled map, {{date}}"
 msgstr "Untitled map, {{date}}"
@@ -416,9 +416,6 @@ msgstr "Line/stroke width"
 msgid "Line/stroke width must be between 0-10."
 msgstr "Line/stroke width must be between 0-10."
 
-msgid "Point size"
-msgstr "Point size"
-
 msgid "Labels"
 msgstr "Labels"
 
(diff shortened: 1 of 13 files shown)


i18n/en.pot
```diff
@@ -636,8 +639,8 @@ msgstr "Group"
 msgid "Filters"
 msgstr "Filters"
 
-msgid "Failed to load layer \"{{layername}}\": {{message}}"
-msgstr "Failed to load layer \"{{layername}}\": {{message}}"
+msgid "{{layername}}: {{message}}"
```
martinkrulltott, line 642:
```suggestion
msgid "{{- layername}}: {{- message}}"
```
is the escaping needed or could it be an issue if the strings contain special characters?

jenniferarnesen: Good point. The message is always the same, so that can be hardcoded. The layername will change though.

src/loaders/geoJsonUrlLoader.js
```diff
@@ -21,49 +26,20 @@ const fetchData = async (url, engine, instanceBaseUrl) => {
...
-                    ) {
-                        throw new Error(i18n.t('Url to geojson was not found.'))
-                    }
-
-                    throw new Error(e.details.message)
-                }
-
-                throw new Error(e)
+            .catch(() => {
+                throw new Error('Error')
```
martinkrulltott, line 30: This looks strange to me. Can you explain how this will be shown to the user? And the purpose of passing `'Error'` (instead of just `new Error()`)

martinkrulltott (approved):

i18n/en.pot
martinkrulltott: Ok, if the `layername` is considered "safe" then I suggest adding `{{- layername}}` to avoid potential escaping issues in the future

PR #3197, opened 2024-05-07
chore: remove fetching of schemas and use the data engine instead for api requests

jenniferarnesen:
Removed all the schemas. 

Where needed, d2 was replaced with data engine from app-runtime.

GroupStyle - change of image path works:
image


janhenrikoverland (approved):
Approving based on @BRaimbault's review.

PR #3232, opened 2024-06-04
refactor: use shared dashboard plugin wrapper

edoardo:
Implements [DHIS2-XXXX](https://dhis2.atlassian.net/browse/DHIS2-XXXX)

**Requires **
---

### Key features

1. use shared dashboard plugin wrapper

---

### Description

Generic components for plugins are available in `app-platform` and `app-runtime`.
Analytics apps are currently having the same components replicated in each repo, a shared dashboard plugin wrapper has been added in `analytics`.

---

### Known issues

-   [ ] an infinite loop happened and [this](https://github.com/dhis2/dashboard-app/commit/03059501e58e2940efa5a750dac099cb612e5c29) seems to solve it, I'm not sure why and if this is a proper solution.

---

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2024-05-07T11:26:24.266Z\n"
-"PO-Revision-Date: 2024-05-07T11:26:24.268Z\n"
+"POT-Creation-Date: 2024-05-30T09:45:16.324Z\n"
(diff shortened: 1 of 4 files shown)


src/components/map/MapLoadingMask.js
```diff
@@ -1,16 +1,10 @@
...
 
 export const loadingMaskClass = 'dhis2-map-loading-mask'
 
 const MapLoadingMask = () => (
-    <ComponentCover
-        translucent
-        className={cx(styles.cover, loadingMaskClass)}
-        dataTest="map-loading-mask"
-    >
+    <ComponentCover className={loadingMaskClass} dataTest="map-loading-mask">
```
jenniferarnesen, line 7: Was removal of `translucent` to fix something else, or was it for these plugin changes?

edoardo: This is to avoid the (one) flashing grey background.

PR #3198, opened 2024-05-08
feat: allow admins to select earth engine layers to show in the app (DHIS2-16098) [ee-dev]

turban:
Implements: https://dhis2.atlassian.net/browse/DHIS2-16098

As we plan add support for additional layers from Google Earth Engine, we should allow system administrators to select which layers that should be visible on their DHIS2 instance. Instead of showing all EE layers in the Add layer dialog, system administrators can select the layers that should be visible from a predefined collection. 

TODO: 
- [x] Get the right authority ID from @janhenrikoverland (should have been included in DHIS2 core 2.41)
- [x] Write some cypress tests
- [x] Gather info about data store and make any relevant changes to this implementation
- [x] Determine whether feature toggling is needed (if 2.40 is supported, then app needs to work with 2.40.0)

With this PR we will show no EE layers in the Add layer dialog by default. Admins having the right authority will see a "Manage available layers"-button: 

Screenshot 2024-05-08 at 17 03 47

When the user clicks this dialog, the supported EE layers will be shown in a modal: 

Screenshot 2024-05-08 at 16 50 48

The visibility of the individual layers can be toggled on and off. The checked layers will now show in the Add layer dialog for all users of the app: 

Screenshot 2024-05-08 at 16 52 09

PR #3245, opened 2024-06-13
fix: reserve namespace, check integrity, use redux

BRaimbault:
A few steps to build upon initial layers types handling via dataStore
- Reserve namespace in app config and changed name to DHIS_WEB_APP to attempt to make it unique (prevent read/write from users without app access, better permission handling is underway: https://dhis2.atlassian.net/browse/DHIS2-17546).
- Used redux store to keep track of layer types visibility setting (avoids requests to datastore beyond the initial one for standard users - also makes the experience a tiny bit smoother -, prevents failures in case of integrity compromized in the middle of a session).
- Added initial step to check data store integrity and re-initialize the key value in case it is compromised (maybe I should add an alert in this scenario?).


BRaimbault:
I am now thinking of a scenario where a lot of users would be given access to the layer types toggling feature. 
With the implementation here the dataStore would be replaced with the selection from each user's session.
I am wondering if this is a scenario we want to dig in more.

d2.config.js
```diff
@@ -20,6 +20,8 @@ const config = {
     },
 
     coreApp: true,
+
+    dataStoreNamespace: 'DHIS_WEB_MAPS',
```
jenniferarnesen, line 24: I'm thinking we could make this namespace even more unique to avoid potential conflicts.

BRaimbault: fix: [455af79](https://github.com/dhis2/maps-app/pull/3198/commits/455af79d46a6b88a0fcb6716db15de1d4f31c068)

PR #3255, opened 2024-06-18
chore: remove orgUnit, orgUnitGroup, program, programStage schemas

jenniferarnesen:
Removed schemas:
* organisationUnit
* organisationUnitGroup
* program
* programStage

Relevant parts of the app to test:
* Data download of an event layer
* Data table with event layer


BRaimbault:
Testing with "Inpatient morbidity and mortality" in https://test.e2e.dhis2.org/analytics-2.40
I see the data table is not showing the data elements marked to "show in report" (Mode of discharge, Age and Diagnosis) and an error is thrown (An unknown error occurred while reading layer data).

Before:
image

After:
image

src/util/event.js
BRaimbault, line 1: Swap `d2` for `engine` when calling `getEventColumns`, that should solve:  https://github.com/dhis2/maps-app/pull/3255#issuecomment-2272955625

BRaimbault (changes requested):

BRaimbault (approved):

PR #3269, opened 2024-06-25
fix: support "week", "bi-week" and months in period translations

BRaimbault:
Part of the fix for [DHIS2-16904](https://jira.dhis2.org/browse/DHIS2-16904)

Requires https://github.com/dhis2/analytics/pull/1685

--------------

## Key features
- bump analytics dependency
- pass periodsSettings to @dhis2/analytics utility function: `getFixedPeriodsOptionsById`

-------------

## Description

The actual fix is in @dhis2/multi-calendar-dates which is a dependency of analytics and it's used in the PeriodSelector component the analytics apps use.
See also the PR in analytics.

For maps-app, the locale and calendar specs where not passed to the @dhis2/analytics utility function, so this was implemented as well.
_Note: this has not been implemented for selectors limited to years._

-------------

## Screenshots
Translations don't work for week/bi-week:
Pending

Before / After

[DHIS2-16904]: https://dhis2.atlassian.net/browse/DHIS2-16904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util/app.js
+++ b/src/util/app.js
@@ -27,6 +27,15 @@ export const appQueries = {
             key: SYSTEM_SETTINGS,
         },
     },
+    systemInfo: {
+        resource: 'system/info',
+    },
+    userSettings: {
+        resource: 'userSettings',
+        params: {
+            key: ['keyUiLocale'],
+        },
+    },
     externalMapLayers: fetchExternalLayersQuery,
 }
 
@@ -70,6 +79,8 @@ const getLayerTypes = (externalMapLayers) => {
 export const providerDataTransformation = ({
     currentUser,
     systemSettings,
+    systemInfo,
+    userSettings,
     externalMapLayers,
 }) => ({
     currentUser: {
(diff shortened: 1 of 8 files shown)


src/util/app.js
```diff
@@ -27,6 +27,15 @@ export const appQueries = {
             key: SYSTEM_SETTINGS,
         },
     },
+    systemInfo: {
```
kabaros, line 30: I think you can get the systemInfo using the [useConfig](https://developers.dhis2.org/docs/app-runtime/hooks/useconfig/) hook - this would make use of the settings already in the provider rather than making an extra API call.

```js
import { useConfig } from '@dhis2/app-runtime'

const { systemInfo } = useConfig()
const { calendar = 'gregory' } = systemInfo
```

kabaros (approved):
lgtm - just the one comment/suggestion

src/util/app.js
```diff
@@ -27,6 +27,15 @@ export const appQueries = {
...
+        resource: 'system/info',
```
edoardo, line 31: Perhaps only request `calendar` to avoid a "bloated" response with information we don't use.

edoardo (approved):

src/util/app.js
BRaimbault: Addressed here: [a594f89](https://github.com/dhis2/maps-app/pull/3269/commits/a594f8963087a3dbdcb0eecdc003027de47558e7)

PR #3286, opened 2024-07-24
fix: avoid crash if layer title is undefined

turban:
To set the `data-test` attribute on the `LayerCard` div we use replace to remove whitespace from the `title` (introduced in https://github.com/dhis2/maps-app/pull/3021).

In some edge cases `title` can be undefined, and we avoid this error if we set the fallback value to `''`. 

This crash bug was reported for https://qa.im.dhis2.org/master/dhis-web-maps/#/AHWtSmx21sx

The error occurs for two saved maps with old EE layers that [should be removed from from Sierra Leone demo database](https://dhis2.atlassian.net/browse/DHIS2-17795): 
- Map: Nighttime lights (AHWtSmx21sx)
- Malaria: Cases with elevation eastern area (gJ1BHisY9Wm)


src/components/layers/LayerCard.js
```diff
@@ -9,7 +9,7 @@ import styles from './styles/LayerCard.module.css'
 import LayerToolbar from './toolbar/LayerToolbar.js'
 
 const LayerCard = ({
-    title,
+    title = '',
```
jenniferarnesen, line 12: Bjørn, should we be concerned that maps saved in this old format don't display? You can see this in 2.39 if you open these same maps. The app doesn't crash (unless you open the Edit layer dialog), but the layer doesn't get displayed.

turban: I think we should remove these maps from the demo db. 
https://dhis2.atlassian.net/browse/DHIS2-17795

jenniferarnesen (approved):

PR #3299, opened 2024-08-07
chore: update jest config

adeldhis2:
update jest config so it only reports to report portal when on CI
default debug to false


adeldhis2:
Updated the Jest config to report to Report Portal only in CI when everything is set up correctly. Also, skipped reporting for Dependabot PRs to prevent errors from missing API keys.

jenniferarnesen (approved):

PR #3302, opened 2024-08-14
fix: plugin crashes for Earth Engine layers in the dashboard

turban:
Maps plugin is crashing in the dasboard for EE maps due to a missing `filters` prop.  It also fixes issues with backward compatibility issues with older maps with EE layers. 

With this fix EE layers load on Dashboard: 

Screenshot 2024-08-14 at 14 04 51

Thematic layers added to the same map respects the dashboard filters: 

Screenshot 2024-08-14 at 14 04 39

### DIFF
--- a/src/loaders/earthEngineLoader.js
+++ b/src/loaders/earthEngineLoader.js
@@ -84,8 +84,8 @@ const earthEngineLoader = async (config) => {
         // From database as favorite
         layerConfig = JSON.parse(config.config)
 
-        // Backward compability for layers with periods saved before 2.36
         if (layerConfig.image) {
+            // Backward compability for layers with periods saved before 2.36
             const filter = layerConfig.filter?.[0]
 
             if (filter) {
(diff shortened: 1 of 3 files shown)


src/loaders/earthEngineLoader.js
```diff
@@ -84,48 +84,38 @@ const earthEngineLoader = async (config) => {
...
+            delete layerConfig.filter
         }
 
-        // Backward compability for layers saved before 2.40
-        if (typeof layerConfig.params?.palette === 'string') {
-            layerConfig.params.palette = layerConfig.params.palette.split(',')
+        // Backward compability for layers saved before v100.6.0
+        if (layerConfig.params) {
+            layerConfig.style = layerConfig.params
+            // Backward compability for layers saved before 2.40
```
jenniferarnesen, line 113: Is this comment correct? Because this block of code splitting palette into an array was added in the big ee refactor PR. My guess is that the comment can be removed since it would actually be the same comment as on line 110?

jenniferarnesen (approved):

PR #3403, opened 2024-11-12
chore: use correct maps bundle name in release workflow

jenniferarnesen:


HendrikThePendric (approved):
Did a `yarn build` in my local repo and got a `.build/bundle/maps-100.7.1.zip`, so this change makes perfect sense: there is no `-app` suffix in the filename.

PR #3470, opened 2025-02-18
fix: "Coordinate field" defaults to "Event location" when "Data" tab is opened

BRaimbault:
Implements [DHIS2-17710](https://dhis2.atlassian.net/browse/DHIS2-17710)

---

### Description

- Prevent setting eventCoordinateField to defaultValue while list of options is incomplete.
- Only set eventCoordinateField to defaultValue if list of options is complete and the previous value cannot be maintained.
- Fetch information to determine if the program is with/without registration (`program.trackedEntityType` is not saved with maps).

---

### Screenshots

image

[DHIS2-17710]: https://dhis2.atlassian.net/browse/DHIS2-17710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BRaimbault:
**1. `useDataQuery`'s `loading`**

`useDataQuery`'s `loading` was always false so using `loading` below was not enough and I switched to `fetching` ([reference](https://github.com/dhis2/app-runtime/blob/master/services/data/src/react/hooks/useDataQuery.ts#L155-L169)), I guess this might be a scenario where "there is already data for the query", but I happy to discuss.

https://github.com/dhis2/maps-app/blob/542eedef555a393569bb2b9a505210273b29d439/src/components/dataItem/CoordinateField.js#L85-L90

**2. Cypress test**

Would you add a cypress test for this?

jenniferarnesen:
Should we just remove `eventCoordinateField` from the CoordinateField component since it isn't in use right now?

jenniferarnesen (approved):

BRaimbault:
> Should we just remove eventCoordinateField from the CoordinateField component since it isn't in use right now?
As discussed, I think we should maintain this part of the code for now given the objective is to enable the associated feature in the near future.

src/components/dataItem/CoordinateField.js
```diff
@@ -23,16 +22,12 @@ const CoordinateField = ({
     onChange,
     className,
 }) => {
-    const { eventDataItems, loading } = useEventDataItems({
+    const { eventDataItems, trackedEntityType, loading } = useEventDataItems({
```
jenniferarnesen, line 25: I'm curious why you changed to get trackedEntityType from here rather than from the program (old code, line 34)

jenniferarnesen (approved):

src/components/dataItem/CoordinateField.js
BRaimbault: Because this is not available in saved maps.

PR #3482, opened 2025-03-07
chore: remove deprecated/removed object references (DHIS2-19062)

edoardo:
Implements [DHIS2-19062](https://dhis2.atlassian.net/browse/DHIS2-19062)

---

### Description

Remove references to deprecated/removed fields `userAccesses`, `userGroupAccesses`, `externalAccess`, `publicAccess`.
The `legendFields` is not used anywhere so it's been removed entirely.

### TODO

- [x] Dashboard tested
- [x] Tests updated (Cypress and/or Jest)

[DHIS2-19062]: https://dhis2.atlassian.net/browse/DHIS2-19062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/helpers.js
```diff
@@ -35,12 +34,10 @@ const analysisFields = () => {
         'trackedEntityType[id,displayName~rename(name)]',
         'organisationUnitSelectionMode',
         '!href',
-        '!publicAccess',
```
jenniferarnesen, line 38: I considered whether we should leave these alone (lines 38, 43, 52) since removing them results in these fields now being returned in the 2.40 maps response. But they are obviously not in use since they're currently being removed, and when saving a map, they are not included in the payload (see change in helpers file)

jenniferarnesen (approved):

jenniferarnesen (approved):

src/util/helpers.js
jenniferarnesen: Decided to leave these exclusions in until v40 is no longer in support.

PR #3488, opened 2025-03-14
fix: multiple improvements to style by data item in event layers [DHIS2-19215]

BRaimbault:
Implements [DHIS2-19215](https://dhis2.atlassian.net/browse/DHIS2-19215)

---

### Description

* "Style by data item" drop down in "Event layer"'s "Style" tab sets to "None" when no styling is applied instead of empty.
image

* Allow changing value if previously selected style data item or filter is not available anymore.
image

* Consider `TRUE_ONLY` a boolean type.
image

* Do not consider data item of type `AGE` numerical (the value stored in DB is a date).
image

* Refactor `src/util/styleByDataItem.js` to:

  * Distinguish between "Other" in legend (not covered by other categories) and "Not set" in popup which indicates a missing value.
image

  * Always handle "Other"/"Not set" properly: across all data types (properly distinguish 0 and no data), with clustered or spread events (screenshot below is just and example, too many scenario). 
image 

  * Always count the number of events associated with each legend item (was only for numerical and excluding "Not set").
image

  * Have a proper default for types not explicitly supported: name of selected data item is visible, count of events is available.
image

  * Use `OptionSetStyle` if the data item is associated with an option set (even if type is numerical or boolean).
image

Reference: https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#metadata-attribute-value-type-and-validations

---

### TODO

- [x] Dashboard tested
- [ ] Tests added (Cypress and/or Jest) N/A
- [ ] Docs added N/A
- [ ] Strings generated N/A
- [ ] d2-ci dependency replaced N/A

[DHIS2-19215]: https://dhis2.atlassian.net/browse/DHIS2-19215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/styleByDataItem.js
```diff
@@ -5,24 +5,29 @@ import {
...
     CLASSIFICATION_PREDEFINED,
 } from '../constants/layers.js'
-import { numberValueTypes } from '../constants/valueTypes.js'
+import { numberValueTypes, booleanValueTypes } from '../constants/valueTypes.js'
 import { cssColor } from '../util/colors.js'
 import { OPTION_SET_QUERY, LEGEND_SET_QUERY } from '../util/requests.js'
 import { getLegendItemForValue } from './classify.js'
 import { getAutomaticLegendItems, getPredefinedLegendItems } from './legend.js'
 
+// Returns true if value is not undefined or null;
```
jenniferarnesen, line 14: No need for the comment - the code is very clear (and also the comment doesn't include the last possibility `value !== ''`

src/util/styleByDataItem.js
```diff
@@ -34,23 +39,51 @@ export const styleByDataItem = async (config, engine) => {
...
-        }
+        const value = feature.properties[id]
 
         return {
             ...feature,
             properties: {
                 ...feature.properties,
-                value: value === '1' ? i18n.t('Yes') : i18n.t('No'),
-                color: value === '1' ? values.true : values.false,
+                value:
```
jenniferarnesen, line 73: Looks like SonarQube isn't set up on this repo, but if it were, it would probably flag this code as raising the complexity score. Is there a more clear way of writing this, even if it is more verbose?

src/components/dataItem/StyleByDataItem.js
```diff
@@ -56,7 +56,7 @@ const StyleByDataItem = ({ program, programStage, error }) => {
         <div>
             <SelectField
                 label={i18n.t('Style by data item')}
-                value={styleDataItem ? styleDataItem.id : null}
+                value={styleDataItem ? styleDataItem.id : 'none'}
```
jenniferarnesen, line 59: The string 'none' is used 2x in this file. Perhaps make this into a constant?

src/util/styleByDataItem.js
BRaimbault: Logic updated to clearly distiguish the 3 cases: true, false and missing/unexpected value:
https://github.com/dhis2/maps-app/pull/3488/commits/30e237ece06ff4a7744b48b1640982aa2da360ca

src/components/dataItem/StyleByDataItem.js
BRaimbault: Replaced by `ITEM_NONE.id`: https://github.com/dhis2/maps-app/pull/3488/commits/30e237ece06ff4a7744b48b1640982aa2da360ca

src/components/dataItem/StyleByDataItem.js
```diff
@@ -38,31 +38,39 @@ const StyleByDataItem = ({ program, programStage, error }) => {
...
-                    name: styleDataItem.name,
-                    nsSeparator: '^^',
-                })}
-            </Help>
+    const internalError =
+        styleDataItem && !dataItems.find((item) => item.id === styleDataItem.id)
+    let internalErrorText
+    if (internalError) {
+        internalErrorText = i18n.t(
+            'Previously selected value not available in list: {{id}}',
```
jenniferarnesen, line 46: Do you think it's useful to the user to show this id (assuming no access to the name)? I suppose the user could report the id to an admin.

What is the scenario where a value is not available?

src/components/dataItem/StyleByDataItem.js
```diff
@@ -38,31 +38,39 @@ const StyleByDataItem = ({ program, programStage, error }) => {
...
     }
 
     return (
         <div>
             <SelectField
                 label={i18n.t('Style by data item')}
-                value={styleDataItem ? styleDataItem.id : null}
+                value={
+                    !(error || internalError)
+                        ? styleDataItem
```
jenniferarnesen, line 60: Could you extract the nested tertiary? I don't think it would pass a Code Climate check if it were enabled on this repo.

src/util/__tests__/styleByDataItem.spec.js
```diff
@@ -0,0 +1,426 @@
...
+                id: STYLE_DATA_ITEM_ID,
+                valueType: numberValueTypes[3],
+            },
+            data: [
+                { properties: { [STYLE_DATA_ITEM_ID]: 0.5 } },
+                { properties: { [STYLE_DATA_ITEM_ID]: 1.5 } },
+                { properties: { [STYLE_DATA_ITEM_ID]: 2.5 } },
+                { properties: { [STYLE_DATA_ITEM_ID]: 3.5 } },
+                { properties: {} },
+                { properties: { [STYLE_DATA_ITEM_ID]: SOME_VALUE } },
```
jenniferarnesen, line 139: Could this actually happen - a numeric value type has a non-numeric value?

src/util/__tests__/styleByDataItem.spec.js
```diff
@@ -0,0 +1,426 @@
...
+        })
+        expect(result.data[2].properties).toMatchObject({
+            value: NOTSET_VALUE,
+            color: EVENT_COLOR,
+        })
+        expect(result.data[3].properties).toMatchObject({
+            value: SOME_VALUE,
+            color: EVENT_COLOR,
+        })
+        expect(result.data[2].properties).toMatchObject({
```
jenniferarnesen, line 394: Looks like lines 394-401 are the same as 386-393?

src/util/__tests__/styleByDataItem.spec.js
```diff
@@ -0,0 +1,426 @@
...
+                    endValue: 3,
+                    color: 'red',
+                    radius: 5,
+                    count: 1,
+                }),
+                expect.objectContaining({
+                    name: LEGEND_ITEM_OTHER,
+                    color: EVENT_COLOR,
+                    radius: 5,
+                    count: 3,
```
jenniferarnesen, line 206: Does this mean that `Not set` and `Other` will grouped together in the legend as `Other`?

src/util/__tests__/styleByDataItem.spec.js
BRaimbault: I am not sure of what happens when you change the data type of a data item which already had values. It is not a good idea for sure, but what will the analytics endpoint return?

src/util/__tests__/styleByDataItem.spec.js
BRaimbault: Thanks, this looks like a copy-paste oversight.

BRaimbault: https://github.com/dhis2/maps-app/pull/3488/commits/b29a7a1adb024015d74aa2e612ae88e9475db981

src/util/__tests__/styleByDataItem.spec.js
BRaimbault: Yes, that is the design.
If you want to check, the popup will indicate whether it is one case or the other.

Originally all was shown as "Not set", it was pointed out by Hendrik during the KFMT that having values styled as "Not set" was misleading.
Distinguishing the two cases properly would require additional specifications/work: we would need another color picker, maybe we should have the possibility to disable it to avoid cluttering the legend, etc.

While I recognize there is room to give more flexibility, I believe the implemented solution is "good enough".

src/util/__tests__/styleByDataItem.spec.js
jenniferarnesen: Seems like an interesting test case.

src/components/dataItem/StyleByDataItem.js
BRaimbault: Here is the summary of the discussion with him:
> Generally speaking, the approach we have for errors is "expose information where it can be useful".
> I think in the case you describe, knowing the offending data item ID is useful - can be reported to an admin or the Map owner. I'd probably change it to the standard Help text color (as it is information, not error).

src/components/map/layers/EventPopup.js
```diff
@@ -23,14 +29,18 @@ const getDataRows = ({ displayItems, dataValues }) => {
     // Include rows for each data item used for styling and displayInReport
     displayItems.forEach(({ id, name, valueType, options }) => {
         const { value } = dataValues.find((d) => d.dataElement === id) || {}
-        let formattedValue = value
 
-        if (valueType === 'COORDINATE' && value) {
-            formattedValue = formatCoordinate(value)
-        } else if (!hasValue(value)) {
+        let formattedValue
```
jenniferarnesen, line 33: Can you extract this into a util function and add tests?

BRaimbault: https://github.com/dhis2/maps-app/pull/3488/commits/16042e26778cd7eef840611de77402bd6b12acc6

jenniferarnesen (approved):

PR #3502, opened 2025-03-25
fix: ensure some d2 init requests are cached

edoardo:
Implements _jira issue link_

---

### Description

Two requests made by d2 didn't end up in the app cache, which prevented the Maps app to work offline.
This is a temporary fix until d2 is completely removed from the app.

### TODO

- [ ] Dashboard tested
- [ ] Tests added (Cypress and/or Jest)
- [ ] Docs added
- [ ] Strings generated
- [ ] d2-ci dependency replaced
- [ ] _todo_

---

### Screenshots

_supporting images_

### DIFF
--- a/d2.config.js
+++ b/d2.config.js
@@ -11,7 +11,9 @@ const config = {
     pwa: {
         enabled: true,
         caching: {
-            patternsToOmitFromAppShell: [/.*/],
+            patternsToOmitFromAppShell: [
+                /^(?!.*me\/authorization)(?!.*me\?fields).*$/,
+            ],
             globsToOmitFromPrecache: ['fonts/**', 'images/**'],
         },
     },


d2.config.js
```diff
@@ -11,7 +11,9 @@ const config = {
     pwa: {
         enabled: true,
         caching: {
-            patternsToOmitFromAppShell: [/.*/],
+            patternsToOmitFromAppShell: [
+                /^(?!.*me\/authorization)(?!.*me\?fields).*$/,
```
BRaimbault, line 15: I needed to add a couple of exceptions to have all types of maps work:
`/^(?!.*me\/authorization)(?!.*me\?fields)(?!.*systemSettings)(?!.*analytics).*$/`
I am not too sure about `analytics` but without it I had only the first map of the "Cases Malaria" loading.

edoardo: I think the RE for `analytics` is a bit too generic here, it would match all requests on our test instances because they contain `analytics` in the path).
If we need to add it we should perhaps use this instead:
```suggestion
                /^(?!.*me\/authorization)(?!.*me\?fields)(?!.*systemSettings)(?!.*analytics\?).*$/,
```

BRaimbault (approved):

PR #3523, opened 2025-05-06
chore: get full version from dhis2 instance in cypress [DHIS2-19491]

BRaimbault:
Implements [DHIS2-19491](https://dhis2.atlassian.net/browse/DHIS2-19491)

### Description

Retrieve full version of dhis2 instance from `api/system/info?fields=version` and store in Cypress env variable `dhis2InstanceFullVersion`.
Give easy access via utility function `getDhis2Version`.

---

### Quality checklist

- [ ] Dashboard tested N/A
- [ ] Cypress and/or Jest tests added/updated N/A
- [ ] Docs added N/A
- [ ] d2-ci dependencies replaced (analytics or maps-gl link https://github.com/dhis2/[lib]/pull/XXX) N/A
- [ ] Tester approved (Hendrik)

[DHIS2-19491]: https://dhis2.atlassian.net/browse/DHIS2-19491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/support/e2e.js
+++ b/cypress/support/e2e.js
@@ -69,7 +69,7 @@ before(() => {
         url: `${Cypress.env('dhis2BaseUrl')}/api/system/info?fields=version`,
     }).then(({ body: { version } }) => {
         const match = version.match(
-            /^(\d+)\.(\d+)\.(\d+)(?:\.(\d+))?(?:-([\w.]+))?$/
+            /^(\d+)\.(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:-([\w.]+))?$/
         )
 
         if (!match) {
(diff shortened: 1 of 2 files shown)


HendrikThePendric (approved):
I think this works well given the current setup.

In a way I find it a bit counter-intuitive that we are finding out what version we are testing during the test, and after we've logged in. But practically speaking I think this is probably the most pragmatic solution...

In theory it could be nice to get the version from our [releases page](https://releases.dhis2.org/v1/versions/stable.json) instead, but since you don't get RCs there, and you'd still have to find the release matching the instance version in the Cypress env variable `dhis2InstanceVersion` and that this instance variable can come in [various flavours](https://github.com/dhis2/dashboard-app/blob/master/cypress/support/utils.js#L61), the solution you've got here is probably your best option.

PR #3545, opened 2025-06-16
chore: replace analytics CachedDataQueryProvider by maps CachedDataProvider

BRaimbault:
### Description

Replace @dhis2/analytics `CachedDataQueryProvider` with maps-app specific `AppDataProvider` that supports async `dataTransformation`.

---

### Quality checklist

- [ ] Dashboard tested
- [ ] Cypress and/or Jest tests added/updated _N/A_
- [ ] Docs added _N/A_
- [ ] d2-ci dependencies replaced (analytics or maps-gl link https://github.com/dhis2/[lib]/pull/XXX) _N/A_
- [x] Tester approved (JJA)

### DIFF
--- a/src/components/plugin/Plugin.js
+++ b/src/components/plugin/Plugin.js
@@ -5,7 +5,7 @@ import {
     SYSTEM_SETTINGS,
 } from '../../constants/settings.js'
 import { getHiddenPeriods } from '../../util/periods.js'
-import { AppDataProvider } from '../app/AppDataProvider.js'
+import { CachedDataProvider } from '../cachedDataProvider/CachedDataProvider.js'
 import MapContainer from './MapContainer.js'
 
 const query = {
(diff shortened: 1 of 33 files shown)


src/components/plugin/Plugin.js
```diff
@@ -50,7 +50,7 @@ const providerDataTransformation = ({ systemSettings, currentUser }) => {
 
 export const Plugin = ({ visualization, displayProperty }) => {
     return (
-        <CachedDataQueryProvider
+        <AppDataProvider
```
jenniferarnesen, line 53: There's an argument for giving this component a different name since it is used by the plugin. I'm not sure what it would be, and it's probably not a big deal. But if you do think of a more generic name, let's go with that instead.

jenniferarnesen: CachedDataProvider

jenniferarnesen:
Does the documentation need to be updated?

jenniferarnesen (changes requested):
See comment

src/components/plugin/Plugin.js
BRaimbault: Makes sense.

BRaimbault: https://github.com/dhis2/maps-app/pull/3545/commits/811efc6d7430db2f50b8c78454aecd4859cd32ba

jenniferarnesen (approved):

jenniferarnesen:
Testing approved by virtue of the e2e tests passing

PR #3652, opened 2026-05-18
docs: split maps.md into per-section source files and refresh content [PR8] [DHIS2-18242]

BRaimbault:
### Summary

The monolithic `docs/maps.md` file (~1900 lines) has been split into numbered source files under `docs/src/`, making individual sections easier to find, edit, and review. A build script reassembles them into the single output file that DHIS2 documentation tooling consumes.

- Split `docs/maps.md` into 20 numbered source files under `docs/src/` (one per section, e.g. `04-thematic-layer.md`, `09-earth-engine-layer.md`)
- Added `scripts/build-docs.js` to concatenate source files back into `docs/maps.md` (with image path rewriting)
- Added `docs:build` and docs:format npm scripts (`yarn docs:build` formats sources then regenerates the output file)
- Updated `.prettierignore` to allow formatting of `docs/src/*.md` while ignoring the generated `docs/maps.md` and `docs/resources/`
- Refreshed all layer dialog screenshots to reflect the current UI
- Reviewed and improved prose across all documentation sections

### Notes

`docs/maps.md` is now a generated file - edits should be made to the source files in `docs/src/` and the output rebuilt with `yarn docs:build`.


BRaimbault (commented):
Ready for review

docs/src/05-event-layer.md
```diff
@@ -0,0 +1,214 @@
+## Manage event layers { #using_maps_event_layer }
+
+The event layer displays the geographical location of events recorded in DHIS2 programs. When events
+have associated point or polygon coordinates, you can drill down from aggregated data in thematic
+layers to the underlying individual events or cases. Alternative coordinate fields also let you plot
```
karolinelien: How can you drill down from thematic layer to event layer?

docs/src/05-event-layer.md
```diff
@@ -0,0 +1,214 @@
...
+
+    If there is only one stage available for the selected program, the stage is automatically
+    selected.
+
+-   Select a value from the **Coordinate field** to determine which positions are displayed on the
+    map. By default, "Event location" is selected. You can also choose "Org unit location".
+    Depending on the selected program, additional options may include "Tracked entity location",
+    "Enrollment location", and coordinate or org unit type data elements or attributes, such as
+    "Household location" or "Referral facility". The number of events represented on the map may
+    vary depending on the selected option and the availability of coordinates. Org units are
```
karolinelien: "Org units with polygons are represented by their centroids"?

docs/src/05-event-layer.md
```diff
@@ -0,0 +1,214 @@
...
+    represented by their centroids.
+
+-   By default, all events with coordinates are shown on the map. Use the **Event status** field to
+    only show events having one status: Active, Completed, Schedule, Overdue, or Skipped.
```
karolinelien: Don't we have only Active and Completed?

docs/src/01-about.md
```diff
@@ -0,0 +1,84 @@
...
+layer types include thematic maps, events, tracked entities, facilities, org unit boundaries, and
+Earth Engine data such as population, elevation, or climate indicators. You can label features,
+search and filter data, and save or share maps with other users, or download them as an image. Saved
+maps can be added to dashboards in the **Dashboard** app. Thematic layers also integrate with the
+**Data Visualizer** app, letting you open the same data as a chart.
+
+> **Note**
+>
+> To use predefined legends in the **Maps** app, you need to create them first in the
+> **Maintenance** app.
```
karolinelien: We should change all references to Maintenance app to Metadata Management App

docs/src/04-thematic-layer.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### 3. Org units
+
+![](../resources/images/maps_thematic_layer_dialog_ORG_UNITS.png)
+
+-   Select the org units you want to include in the layer. It is possible to select either
+
+    -   One or more specific org units, org unit levels in the hierarchy, org unit groups, or
+
+    -   A relative level in the org unit hierarchy, with respect to the user. By selecting a **User
```
karolinelien: You can combine the level/org unit groups with user org unit, right?

I also tried to rewrite it slightly, rather than "users at different levels"... Let me know what you think 

"...the map data will appear according to the assigned org units for each user in the org unit hierarchy."

karolinelien (commented):
Great job, I added some comments!

docs/src/05-event-layer.md
```diff
@@ -0,0 +1,214 @@
...
+    Multiple filters may be added. Click the trash button on the right of the filter to remove it.
+
+#### 5. Style
+
+![](../resources/images/maps_event_layer_dialog_STYLE.png)
+
+-   Select **Group events** to group nearby events (cluster), or **View all events** to display
+    events individually.
+
+    -   Select a **color** and **radius** (between 1 and 20) for the event or cluster points.
```
karolinelien: 0 and 100

docs/src/06-tracked-entity-layer.md
```diff
@@ -0,0 +1,109 @@
...
+
+    -   **Selected and all below**: Include tracked entities in and all below selected org units.
+
+#### 5. Style
+
+![](../resources/images/maps_tracked_entity_layer_dialog_STYLE.png)
+
+-   Select a **Color** for tracked entity points and polygons.
+
+-   Select the **Point size** (radius between 1 and 20) for the points.
```
karolinelien: Point size actually don't have a max. Should we add one?

(Related entity has max 100)

docs/src/05-event-layer.md
BRaimbault: You are right, the formulation is misleading, you have to manually create the layer.
Updating to:

> The event layer displays the geographical location of events recorded in DHIS2 programs. While in thematic layers you can visualize event data aggregated to org unit level, with event layers you can display the individual events or cases when they have their own point or polygon coordinates. Alternative coordinate fields also let you plot events using locations other than the event itself, such as the org unit or enrollment location.

docs/src/05-event-layer.md
BRaimbault: Updating to:

> When org unit location is used, polygons are represented by their centroids.

docs/src/01-about.md
BRaimbault: Done, only keeping "Maintenance app" when we talking about external layers until we support this within maps app.

docs/src/04-thematic-layer.md
BRaimbault: > -   Select the org units you want to include in the layer. You can freely combine specific org
    units, org unit levels, org unit groups, and relative user org units (**User organisation
    unit**, **User sub-units**, **User sub-x2-units**). When user org units are selected, the map
    data will appear according to the assigned org units for each user in the org unit hierarchy.

    > **Tip**: Levels and groups act as filters within the org units you select — combining them
    > will show only org units at those levels or in those groups that are descendants of your
    > selected org units.

docs/src/06-tracked-entity-layer.md
BRaimbault: Removing the mention.
Note that the max is not actually enforced when typing.

docs/src/05-event-layer.md
BRaimbault: Just removing the mention.

docs/src/05-event-layer.md
BRaimbault: That's correct. Updated.
I think we could support Schedule in the future, these events do not have event coordinates, but we could use other coordinates. Also we could support other dates: schedule date etc..

PR #3669, opened 2026-06-11
fix: render calculated integer values without decimals in timeline [DHIS2-21037]

BRaimbault:
_Re-created from: https://github.com/dhis2/maps-app/pull/3656 after repo rule change on signed commit._

Analytics returns row values as strings, and computed data items (indicators, calculations) serialize whole numbers with a trailing ".0" (e.g. "16082.0"). The single-map path already parseFloats the value, but the timeline / split-by-period path (getValuesByPeriod) kept the raw string, so popups and hover labels rendered the trailing decimal for values that should display as integers.

Coerce the period value with parseFloat to mirror getValueMapsById. This is lossless (single-map display and the GeoJSON export already use numbers) and additionally routes timeline value filtering through the numeric filter instead of substring matching.

AI Assisted

Implements [DHIS2-21037](https://dhis2.atlassian.net/browse/DHIS2-21037)

  ## Test plan

  - [x] New `thematicLoader.spec.js` asserts both extractors yield numbers and strip the trailing `.0`
  - [x] Existing `useTableData.spec.jsx` still passes
  - [x] Manual: Calculation (difference of two integers) as a thematic layer → Timeline / Split by period → confirm integers in popups, hover labels, legend

---

### Quality checklist

- [x] Dashboard tested
- [ ] Docs added _N/A_
- [x] Tester approved (BR)

---
Before:
image

After: 
image

[DHIS2-21037]: https://dhis2.atlassian.net/browse/DHIS2-21037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BRaimbault (commented):
Approved

karolinelien (approved):

PR #3681, opened 2026-06-18
chore: refactor TrackedEntityDialog to functional component [DHIS2-18209]

BRaimbault:
Implements [DHIS2-18209](https://dhis2.atlassian.net/browse/DHIS2-18209)

### Description

Refactor TrackedEntityDialog to functional component.

#### Implementation notes

- **Action creators are now dispatched via `useDispatch()` instead of being prop-injected through `connect()`** — this matches the pattern already established in `FacilityDialog`, `GeoJsonDialog`, and `ThematicDialog`. No behavior change, just the form `dispatch(setX(val))` instead of `setX(val)`.
- **`orgUnits` appears in a `useEffect` dependency array** (TrackedEntityDialog's default-org-unit-roots effect). `OrgUnitsProvider`'s context value isn't memoized, so this effect can re-fire while the dialog is open and `rows` is still empty. Same pattern already ships in `EarthEngineDialog` (`useEffect(..., [rows, orgUnits, setOrgUnits])`), guarded by the same `!rows` check, so this isn't new risk — flagging in case it comes up in review.

---

### Quality checklist

- [ ] Dashboard tested _N/A_
- [ ] Cypress and/or Jest tests added/updated _N/A_
- [ ] Docs added _N/A_
- [x] Tester approved (@edoardo )

[DHIS2-18209]: https://dhis2.atlassian.net/browse/DHIS2-18209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/edit/trackedEntity/TrackedEntityDialog.jsx
```diff
@@ -53,427 +52,362 @@ import PeriodTypeSelect from './PeriodTypeSelect.jsx'
...
-            this.setErrorState('periodError', null, 'period')
+            setPeriodError(undefined)
         }
+    }, [periodError, startDate, prevStartDate, endDate, prevEndDate])
 
-        if (!program && prev.program !== program) {
-            setPeriodType(LAST_UPDATED_DATES)
+    // Reset period type when program is cleared
+    useEffect(() => {
+        if (prevProgram && !program) {
```
edoardo, line 159: Does `prevProgram` always have a value?
The old condition was also checking if `prevProgram !== program`.

edoardo (dismissed):

src/components/edit/trackedEntity/TrackedEntityDialog.jsx
BRaimbault: `prevProgram` is `undefined` on the first render (that's what `usePrevious` returns before any update). When a user selects a TE type on a new layer, `program` transitions from `undefined` to `null`, the original condition `!program && prev.program !== program` would fire here and dispatch `setPeriodType(LAST_UPDATED_DATES)` unnecessarily (the period type is already at its default for a new layer). The new condition `prevProgram && !program` correctly skips this case, only firing when a program was genuinely selected and then cleared.

edoardo (dismissed):

edoardo (approved):

PR #3685, opened 2026-06-18
fix: detect fullscreen and resize to autopin legend and enable multitouch [DHIS2-13242] [DHIS2-20761]

BRaimbault:
Implements [DHIS2-20761](https://dhis2.atlassian.net/browse/DHIS2-20761) and [DHIS2-13242](https://dhis2.atlassian.net/browse/DHIS2-13242)

### Description

- `plugin/Map.jsx`: track real fullscreen state via `parent.document.fullscreenElement` (was hardcoded `false`) and forward it to `MapView`/`Legend`
- `Legend.jsx`: auto-open/pin on fullscreen enter, auto-close/unpin on exit (still manually un-pinnable)
- `util/map.js`: `onFullscreenChange` now fits bounds on both enter and exit (was exit-only), via double-rAF; extracted shared `resizeAndFitBounds` helper
- `Map.jsx`: replaced the leaky/broken window-resize listener with a safe one (resize + fitBounds only, no toggle side-effects) to restore dashboard-slideshow fitBounds, with proper cleanup
- `SplitView.jsx` / `MapItem.jsx`: forward `resizeCount` so split-view panes actually resize on dashboard-triggered resizes (previously silently dropped)
- Added unit tests for `util/map.js` and `plugin/Legend.jsx`

---

### Quality checklist

- [x] Dashboard tested
- [ ] Docs added _N/A_
- [ ] Tester approved (name)

---

### Testing

Bundle: [maps-999.99.9-fullscreen-resize-rc1.2026-06-18.zip](https://github.com/user-attachments/files/29096452/maps-999.99.9-fullscreen-resize-rc1.2026-06-18.zip)

Bundle temporarily deployed to: https://dev.im.dhis2.org/maps-app-42-3

Test dashboard: https://dev.im.dhis2.org/maps-app-42-3/apps/dashboard#/MPNFJ5MXcg8

- In desktop: 
  - Open slideshow, navigate between maps, confirm the legend is open and the layer fill the available space.
  - Close slideshow, confirm the legends are closed and the layer fill the available space.
  - Confirm the map resize properly when the window/map canvas dimension change.
- In mobile:
  - Confirm multitouch is enabled in fullscreen, disabled in normal mode. 

[DHIS2-20761]: https://dhis2.atlassian.net/browse/DHIS2-20761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-13242]: https://dhis2.atlassian.net/browse/DHIS2-13242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric (approved):

HendrikThePendric (approved):
Sorry I did approve before, but I probably should have just left a comment. I don't think there is anything wrong with this PR, but is one important part I actually don't understand...

You've added a `resizeCount` prop to `MapItem` and `SplitView`. In the diff I see you now pass `resizeCount` from `SplitView` to `MapItem`, so that's OK. But then I would also expect to see a similar diff somewhere else in this PR, where a `<SplitView />` is rendered and it getting an new `resizeCount` prop. I never see this though and that's a bit of a mystery to me.

src/components/plugin/Map.jsx
```diff
@@ -25,6 +25,14 @@ const defaultBounds = [
     [50.2, 35.9],
 ]
 
+const getFullscreenDoc = () => {
+    try {
+        return parent.document
+    } catch {
+        return document
+    }
+}
```
HendrikThePendric, line 34: Instead of a try/catch, would this not work?
```suggestion
const getFullscreenDoc = () =>
    parent?.document ?? document
```

BRaimbault:
@HendrikThePendric MapView.jsx already passed `resizeCount={resizeCount}` into `<SplitView>` before this PR.

https://github.com/dhis2/maps-app/blob/b1e033b68e8799d747f2d7bbe050e5feec0c4afc/src/components/map/MapView.jsx#L53-L65

src/components/plugin/Map.jsx
BRaimbault: This is to avoid a crash for the case where the dashboard and the plugin are served from different origins (local development), the browser's Same-Origin Policy blocks reading `parent.document` and throws a `SecurityError`.

HendrikThePendric: Aha! 👍

PR #3697, opened 2026-06-23
fix: correct time period filtering in Earth Engine layers

karolinelien:
_No Jira ticket — found during a repository bug-hunt._

### Description

Fixes a bug where a saved Earth Engine layer with a pre-v100.6.0 backward-compat config (a dynamic `filter` array and **no** `image`) never loads — the user sees a permanent spinner with no error.

In `src/loaders/earthEngineLoader.js`, the `else if (layerConfig.filter)` branch passed a bare `filter` to `getPeriodFromFilter(...)`. That `filter` binding is `const`-declared only inside the sibling `if (layerConfig.image)` block, so in the ES module strict scope the reference threw before initialization. Because the throw happens **outside** the loader's `try/catch`, the loader promise rejected and the layer silently never loaded.

The fix passes `layerConfig.filter` — the whole array, which is what `getPeriodFromFilter(filter, datasetId)` expects (it calls `Array.isArray(filter)` / `filter[0]`). This also corrects the argument *shape*: the old `filter` would have been a single element, not the array.

Added Jest coverage for the affected backward-compat branches:
- a filter-only config (pre-v100.6.0) loads without throwing and derives the `period`
- the period id falls back to the filter `arguments` when no explicit `id` is set
- the sibling pre-2.36 `image` + `filter` path reconstructs the `period`

Verified the new regression test fails against the un-fixed code (`ReferenceError: Cannot access 'filter' before initialization`) and passes with the fix.

---

### Quality checklist

- [ ] Dashboard tested _N/A_
- [ ] Docs added _N/A_
- [ ] d2-ci dependencies replaced — _N/A_
- [x] Tester approved (BR)

---

AI Assisted

### DIFF
--- a/d2.config.js
+++ b/d2.config.js
@@ -1,3 +1,5 @@
+const fs = require('fs')
+
 const omitPatterns = [
     // User info
     'me/authorization',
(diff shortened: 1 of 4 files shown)


BRaimbault (dismissed):
Also adding a fix to make ee-worker work in dev mode.

BRaimbault (approved):

PR #3699, opened 2026-06-23
fix: prevent data table from crashing when cursor leaves browser window

karolinelien:
_No Jira ticket — found during a repository bug-hunt._

### Summary

The data table's `onMouseLeave` handler read `nextElement.tagName` without a null check. When the cursor exits the browser window while hovering a row, `event.relatedTarget` is `null` (and the non-standard `event.toElement` is `undefined` in Firefox), so `nextElement` is `null` and `null.tagName` throws a `TypeError`. The surrounding `ErrorBoundary` then replaces the whole table with "Something went wrong."

Fix: guard the access with optional chaining (`nextElement?.tagName`). When leaving to no element, this evaluates to `undefined !== 'TD'` → `true`, so the highlight is cleared correctly instead of crashing.

### Steps to reproduce (before fix)

1. Open a map layer and open the Data Table
2. Hover over a table row
3. Move the cursor straight out of the browser window (most reliable in Firefox)
4. The table is replaced with "Something went wrong."

### Changes

- `src/components/datatable/DataTable.jsx`: `nextElement.tagName` → `nextElement?.tagName`

---

AI Assisted

### DIFF
--- a/src/components/datatable/DataTable.jsx
+++ b/src/components/datatable/DataTable.jsx
@@ -178,7 +178,7 @@ const Table = ({ availableWidth }) => {
             // When hovering to the next row the next element is a `TD`
             // If this is the case `setFeatureHighlight` will
             // fire and the highlight does not need to be cleared
-            if (nextElement.tagName !== 'TD') {
+            if (nextElement?.tagName !== 'TD') {
                 dispatch(highlightFeature(null))
             }
         },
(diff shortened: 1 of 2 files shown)


BRaimbault (approved):
I cannot reproduce the issue but the change is theoreticaly sound and safe.

PR #3714, opened 2026-07-06
feat: redesign data table toolbar and add row selection context menu [PR1]

BRaimbault:

### Description

Replaces the 20px data table control strip with a full 36px toolbar and adds a right-click context menu to table rows. First of a series of PRs improving the data table experience.

**Toolbar (`BottomPanel`)**
- Shows the active layer name and live row count (`X of Y rows` when filters are active, `X rows` otherwise)
- "Clear filters" button appears when any column filter is active
- Sort arrows in column headers now correctly reflect ascending/descending direction

**Row context menu (`TableContextMenu`)**
- Right-click a row: drill up/down, view profile, zoom to feature
- Drill items are hidden for boundary, facility, event, and GeoJSON URL layers
- "View profile" opens the feature profile for GeoJSON URL layers and the org unit profile for all others

**GeoJSON URL layer**
- Right-click on map features now opens the context menu (was wired only for click)
- Map context menu: "View profile" opens the feature profile; drill up/down items are hidden
- Shared `getGeojsonFeatureProfile` utility keeps the profile format consistent across map click, map context menu, and table context menu

**Bug fixes**
- Feature highlight no longer lost after async layer rebuilds (thematic timeline / split-by-period, EarthEngine full rebuild path)
- Fixed null crash in `highlightFeature` when the map layer has no highlight method
- O(1) feature lookup by id in the data table (was an O(n) scan per interaction)

---

### Quality checklist

- [ ] Dashboard tested _N/A_
- [ ] Docs added _N/A_

---

### Screenshots

_supporting images_


src/components/datatable/BottomPanel.jsx
```diff
@@ -65,6 +103,17 @@ const BottomPanel = () => {
...
+    let rowCountLabel = null
+    if (totalCount !== null && filteredCount !== null) {
+        rowCountLabel =
+            filteredCount < totalCount
+                ? i18n.t('{{filtered}} of {{total}} rows', {
+                      filtered: filteredCount,
+                      total: totalCount,
+                  })
+                : i18n.t('{{total}} rows', { total: totalCount })
+    }
```
HendrikThePendric, line 115: MINOR: Could be nice to wrap in useMemo, not because this is an expensive calculation, but simply because you can then use a `const` instead of `let` and return `null` early.

src/components/datatable/BottomPanel.jsx
```diff
@@ -77,16 +126,66 @@ const BottomPanel = () => {
...
+                        <div
+                            className={styles.nameTooltip}
+                            style={{
+                                top: nameTooltipPos.top,
+                                left: nameTooltipPos.left,
+                                color: nameTooltipPos.color,
+                                fontSize: nameTooltipPos.fontSize,
+                                lineHeight: nameTooltipPos.lineHeight,
+                                paddingLeft: nameTooltipPos.paddingLeft,
+                            }}
```
HendrikThePendric, line 148: MINOR: You can simplify this, provided there will never be any non-style fields.
```suggestion
                            style={nameTooltipPos}
```

Works for `null` too

src/components/datatable/TableContextMenu.jsx
```diff
@@ -0,0 +1,178 @@
...
+        grandParentId,
+        grandParentParentGraph,
+        parentGraph,
+    } = featureProps || {}
+
+    const canDrill =
+        layerType !== BOUNDARY_LAYER &&
+        layerType !== FACILITY_LAYER &&
+        layerType !== EVENT_LAYER &&
+        layerType !== GEOJSON_URL_LAYER
```
HendrikThePendric, line 55: MINOR: Maybe store these layers in a Set in the module scope and do sth like `!UNDRILLABLE_LAYERS.has(layer)`

HendrikThePendric (dismissed):
Left some minor suggestions, but this is looking very tidy 🚀

src/components/datatable/TableContextMenu.jsx
BRaimbault: https://github.com/dhis2/maps-app/pull/3714/commits/45f89bdc902431479aba8bdc643117ff85a0974c

src/components/datatable/BottomPanel.jsx
BRaimbault: https://github.com/dhis2/maps-app/pull/3714/commits/45f89bdc902431479aba8bdc643117ff85a0974c

src/components/datatable/BottomPanel.jsx
BRaimbault: https://github.com/dhis2/maps-app/pull/3714/commits/45f89bdc902431479aba8bdc643117ff85a0974c

PR #3715, opened 2026-07-07
feat: period selector for events layer

BRaimbault:
### Description

Brings the Events layer dialog's Period tab in line with the Thematic layer dialog (Stage 1 of 2 — Timeline/Split view will be Stage 2):

1. **Multi-select periods, or start/end dates** — the Period tab now lets you choose "Choose from presets" (select one or more fixed/relative periods, combined into one query) or "Define start-end dates", replacing the old single-period dropdown. Existing saved Event layers keep working unchanged.
2. **Shared with Thematic** — the period tab's initialization, validation, and request-building logic is now shared between the Events and Thematic dialogs instead of being duplicated.
3. **Cleanup** — removes the old single-period selector and related code now that Events uses the same multi-period model as Thematic.

---

### Quality checklist

- [x] Dashboard tested
- [x] Docs added
- [x] Tester approved (@edoardo )

---

### Screenshots

- Relative and Fixed periods: Before/After
  image

- Start and End dates: Before/After
  image


src/components/edit/event/EventDialog.jsx
```diff
@@ -358,29 +342,67 @@ const EventDialog = ({
...
+                                            dispatch(setPeriods(items))
+                                        }
+                                        excludedPeriodTypes={
+                                            systemSettings.hiddenPeriods
+                                        }
+                                        height="324px"
+                                    />
+                                )}
+                                {periodType === START_END_DATES && (
+                                    <StartEndDate
```
edoardo, line 389: Not in scope for the PR but it seems this component can be extracted and shared via `analytics`.
We have a similar one in other analytics apps.

edoardo (approved):

src/components/edit/event/EventDialog.jsx
BRaimbault: We have a custom date picker in maps-app. 
Agreed we should look at sharing component down the road.

PR #3721, opened 2026-07-13
feat: add multi-select filters and global search across data table rows [PR3]

BRaimbault:

> **Diff size note:** this PR adds ~4,150 lines, ~2,000 are runtime code / ~2,150 (52%) are tests.

### Description

Redesigns the data table's per-column filters into one unified multi-select widget, adds a global search box, and replaces the boolean "show only selected" toggle with a richer selected/not-selected filter. Third in the data table improvement series, building on PR1's toolbar and PR2's selection sync.

**Column filters (`FilterInput`)**
- Every column now gets the same searchable, multi-select popover: the trigger doubles as a live search box, narrowing a virtualized checkbox list of the column's actual distinct values
- Checkboxes support multi-select (OR) filtering, e.g. "High" + "Low" on the Legend column
- Typed text that doesn't match an existing value applies live as a custom filter ("Contains" for text columns, "Use filter" for numeric columns, reusing the existing `>`/`>=`/`<`/`<=`/`,`/`&` syntax); a help tooltip explains the numeric syntax
- Pinned "Any value" (matches every non-blank value) and "No value" (blank-cell sentinel) options, kept above a divider so they stay visible while searching; "Any value" acts as a select-all/deselect-all for the column's real values, kept in sync both ways
- New "Reverse selection" button per column, inverting that column's full value domain (ignores the current search-narrowed list)
- Dropdown sort order now matches the column's active sort direction; the Range column's dropdown sorts numerically instead of lexically
- Sentinel values (blank/"any value"/"selected row") consolidated into `constants/dataTable.js` (`SENTINEL_NO_VALUE`, `SENTINEL_ANY_VALUE`, `SENTINEL_SELECTED_ROW`)

**Global search**
- New "Search all columns" input in the toolbar, case-insensitive substring match across every string column (`filterByGlobalSearch`), combined with column filters and the selection filter

**Selection filter (replaces the boolean "show only selected" toggle)**
- The checkbox column header now has its own filter popover ("Selected" / "Not selected", either/both/neither), synced to the map the same way column filters are (`Layer.js`'s `getVisibleIds`)
- Header "select all" checkbox and the new "Reverse selection" button next to it both only ever touch currently visible/filtered rows — anything selected outside the current filter is left untouched; both now have clearer tooltips ("Select all visible rows" / "Reverse selection of visible rows")

**"Clear filters"**
- Now resets column filters, the selection filter, global search, and "show only in view" together, and is always visible (disabled instead of hidden) so it's discoverable

**Toolbar polish**
- Vertical dividers group related icons; the highlight color picker moved next to the layer name
- Collapse/restore icon simplified from double chevrons to a single chevron
- Consistent top-side tooltip placement across toolbar controls

**Resize handle**
- Rewritten from HTML5 native drag-and-drop to Pointer Events — fixes a "not-allowed" cursor during drag and a cursor that never switched to "grabbing"
- Restyled to match the drag-handle look used elsewhere in the app (e.g. layer cards)

**Context menu**
- "Zoom to filtered features" added to both the map and table context menus, disabled when nothing's currently filtered

**Removed**
- The synthetic "Index" column — it was never a real feature property, only an internal row-order tiebreaker, and filtering by it made map layers disappear entirely (an `index` filter never matched any real feature)

**Refactors (no behavior change)**
- `FilterInput.jsx` split: dropdown-positioning logic → `FilterDropdownPopover.jsx` (also removes a cross-component dependency, since `DataTable.jsx` was reaching into `FilterInput.jsx`'s internals for it); the sentinel-value toggle/reverse logic → `util/filterSelection.js`, now directly unit-tested
- `DataTable.jsx` split: column-width measurement/resize logic → `useColumnWidths.js`; row-selection bookkeeping → `useRowSelection.js`
- `useTableData.js`'s sort/compare logic → `util/tableSort.js`, now directly unit-tested instead of only covered indirectly through the hook
- Sort-direction and DHIS2 CSS token usage consolidated (`SORT_ASCENDING`/`SORT_DESCENDING` in `constants/dataTable.js`; `--colors-white`/`--elevations-popover` custom properties instead of hardcoded values in the datatable stylesheets)

**Bug fixes**
- 5 Cypress tests updated for the new UI (checkbox column shifted other column indices; filter input `data-test` renamed from `-input-` to `-search-`)
- Text no longer overflows unbroken in table cells
- Fixed a crash in the map context menu's `mapStateToProps` when no context menu is open

---

### Quality checklist

- [ ] Docs added _N/A_

---

### Screenshots

_supporting images_


src/components/datatable/BottomPanel.jsx
HendrikThePendric, line 1: I get that `globalSearch` applies a client-side filter, but I am still slightly concerned that there is no debounce, or at least a separation between the input value and the global search value. For the input you just want to apply the value from the change handler directly, but for the search value isn't it better to have a small sanitisation layer, i.e. `str.trim().toLocaleLowerCase()`? You could look into having a `searchInputValue` and a debounced+sanitized globalSearch. Not sure if you have a hooks-helper dependency with a `useDebouncedValue` member. If you'd have to implement it from scratch it's maybe too much work.

src/components/datatable/useColumnWidths.js
```diff
@@ -0,0 +1,66 @@
...
+            const frameId = requestAnimationFrame(() => {
+                if (!headerRowRef.current) {
+                    return
+                }
+
+                const measuredColumnWidths = []
+
+                const dataCells = Array.from(headerRowRef.current.cells).slice(
+                    1
+                )
```
HendrikThePendric, line 22: Why is the first column in the row being excluded? Because it needs a fixed width?

HendrikThePendric (dismissed):
There was a lot of code to review 😅 , but it looks good to me. I left an optional suggestion and a question in the diff.

I also spent a long time re-reading the `src/components/datatable/FilterInput.jsx` file and its surrounding components. There are two issues with the code here as I see it:
- You end up almost re-implementing the tooltip component from @dhis2/ui, with some slightly different behaviour
- The code is quite hard to follow because the tooltip is not able to determine its own positioning strategy

But in the end I couldn't really come up with a better way to approach this. I considered that perhaps it would be better to expose the modifiers array on the Tooltip, so that you could then use the renderProps strategy to stick to using it. But adding such a low-level prop to a higher-level component is problematic. And the fact that the tootlip position needs to be established higher up in the component hierarchy seem intrinsic to the problem you're solving here. Since I can't suggest a better alternative, I should assume that what you have here is optimal.

src/components/datatable/useColumnWidths.js
BRaimbault: Correct.
I added a comment.
https://github.com/dhis2/maps-app/pull/3721/commits/0168ca37e1c405a439f352d382ad84486ab4fd34

src/components/datatable/BottomPanel.jsx
BRaimbault: - Added useDebouncedValue(value, delay)
- Split state into searchInputValue (bound to the input) and debounced globalSearch (200ms) passed to the table
- Sanitisation already happens in filterByGlobalSearch (trim + lowercase), so didn't duplicate it

https://github.com/dhis2/maps-app/pull/3721/commits/0168ca37e1c405a439f352d382ad84486ab4fd34

BRaimbault:
Thanks for your review @HendrikThePendric, I addressed your specific comments and marked them as resolved. Thanks for checking closely the tooltip placement logic, seems overcomplicated, but I think it is worth to keep the UI readable / predictable. I'm leaving it as is.

PR #3724, opened 2026-07-16
feat: add column show/hide, pinning, and reordering to data table [PR4]

BRaimbault:

> **Diff size note:** this PR adds ~3,300 lines: ~2,250 are runtime code / ~1,050 (32%) are tests.

### Description

Adds a column picker to the data table toolbar for showing/hiding, pinning (freeze left), and reordering columns, and persists the configuration per layer in the map's saved favourite. Fourth in the data table improvement series, building on PR1's toolbar, PR2's selection sync, and PR3's filtering.

**Column picker (`ColumnPicker`)**

-   New "Configure columns" toolbar icon opens a popover listing every column for the active layer, each with a checkbox (show/hide), a drag handle (reorder, via `@dnd-kit`, consistent with the drag-reorder already used in `LayersPanel.jsx`), and a pin button (freeze to the left of the table)
-   Pinned columns are grouped at the top of the list with a divider separating them from the rest; the pin icon uses this app's teal "active" accent when a column is pinned
-   Per-control tooltips (drag handle, pin button) replace a persistent hint line; the drag preview is portaled to `document.body` so it renders above the rest of the UI instead of being clipped by the popover's own positioned ancestor
-   Long column names truncate with an ellipsis instead of wrapping and breaking the row layout
-   A bulk-actions row above the list adds a checkbox to show/hide every column at once, a reverse-selection button (the same `IconSync16` already used elsewhere in the table), and a reset-to-defaults button that clears the saved config back to the default order/visibility
-   A search box appears once the column list is long enough to actually scroll, filtering by name without ever resizing the popover
-   The popover's width is measured once from its content when it opens (the widest column label) and then held fixed, so searching/filtering never resizes it

**Pinned columns in the table**

-   Pinned columns render as sticky (`fixed`/`left`) cells using `@dhis2/ui`'s built-in support, staying in place while the rest of the table scrolls horizontally
-   The row-selection checkbox column only becomes sticky once something else is pinned, avoiding an unconditional `<th>`-vs-`<td>` DOM shape change for the common nothing-pinned case

**Persistence**

-   New `dataTableColumnConfig` field on each `mapView`, following the same `config` JSON-blob pattern already used for `legendDecimalPlaces`/`legendIsolated`/etc.: `DATA_TABLE_COLUMN_CONFIG_SET` reducer action, round-tripped through `favorites.js` and every layer loader (thematic, event, org unit, facility, tracked entity, GeoJSON URL, Earth Engine) so column visibility/order/pinning survives reopening or sharing a saved map

**Filter popover polish**

-   Multi-select filter dropdown width is now driven by its own content (the widest option label, measured directly since the list renders through a virtualized `react-virtuoso` component that CSS's own intrinsic sizing can't see through) instead of the anchor column's width, so a wide column no longer forces an oversized dropdown
-   Fixed a layout bug where option rows used a collapsing CSS margin for spacing, understating the list's real height and leaving a visible gap before the "Any value" row
-   "Reverse selection" button gained a missing hover state and a fix for its icon being compressed under the checkbox column's tight layout

**Toolbar polish**

-   Double-clicking the toolbar to collapse/restore the panel no longer fires when the click lands on one of the toolbar's own controls (icon buttons, the color picker, the search field)
-   Sort icons are top-aligned against wrapped column headers instead of centering against the full wrapped label; the active sort arrow and the row-selection checkboxes now use this app's teal accent for consistency with the pin icon
-   Each toolbar control (search, row count, show-in-view toggle, clear filters, highlight color, active layer name, resize handle, close, collapse, column picker) is now its own file under `datatable/controls/`, replacing one large block of inline JSX in `BottomPanel.jsx`

**Bug fixes**

-   Fixed the column picker's trigger button being permanently disabled for layer types with synchronous header computation (Thematic/OrgUnit/Facility) - a child-before-parent React effect-ordering race meant the picker's header state was cleared right after being set; Event layers happened to avoid it only because their extended-events loading triggers a second, later header recompute

**Refactors (no behavior change)**

-   Column visibility/order/pin computation (`getVisibleHeaders`/`getPinnedLeftOffsets`) extracted into `util/tableColumns.js`, directly unit-tested, matching the `filterSelection.js`/`tableSort.js` pattern from PR3
-   `FilterInput.jsx`'s pure filter/search/display-value logic and canvas text measurement extracted into `util/filterInput.js` (unit-tested); its self-contained help tooltip extracted into `FilterHelpTooltip.jsx`
-   `DataTable.jsx`'s self-contained pieces (`TopTooltip`, `SelectionFilterButton`, the `TableVirtuoso` "components" glue) extracted into their own files; its pure row/sort/filterable-column helpers moved into `util/dataTable.js`
-   `ColumnPickerControl.jsx`'s row rendering extracted into `ColumnRow.jsx`; its column-key toggle/pin/reverse-selection logic and pinned-cell layout math moved into `util/tableColumns.js` alongside the existing column helpers - this also removed a duplicated pinned-column-count calculation that had crept into `DataTable.jsx`
-   A couple of values shared between CSS and JS (a tooltip's vertical padding, the toolbar's collapsed height) consolidated into CSS custom properties read via the existing `getCssVar` helper, instead of hand-maintained duplicate constants

---

### Quality checklist

- [ ] Docs added _N/A_

---

### Screenshots

_supporting images_

[DHIS2-10655]: https://dhis2.atlassian.net/browse/DHIS2-10655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/datatable/controls/ActiveLayerControl.jsx
```diff
@@ -0,0 +1,71 @@
...
+        if (!el || el.scrollWidth <= el.offsetWidth) {
+            return
+        }
+        const rect = el.getBoundingClientRect()
+        const computed = getComputedStyle(el)
+        const lineHeight = Number.parseFloat(computed.lineHeight)
+        const verticalPadding = getCssVar(
+            '--data-table-name-tooltip-vertical-padding'
+        )
+        setNameTooltipPos({
```
edoardo: I guess `Pos` here refers to `position`.
The object has more than that, perhaps rename to `setNameTooltipProps`.

src/components/datatable/controls/ToolbarIconButton.jsx
```diff
@@ -0,0 +1,40 @@
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import React, { forwardRef } from 'react'
+import IconButton from '../../core/IconButton.jsx'
+import styles from './styles/ToolbarIconButton.module.css'
+
+const ToolbarIconButton = forwardRef(
```
edoardo, line 7: Is this needed since it's basically the same as `IconButton` except for the `active` prop?

src/components/datatable/DataTable.jsx
```diff
@@ -425,12 +179,40 @@ const Table = ({
         globalSearch,
     })
 
+    useEffect(() => {
+        onHeadersChange?.(headers)
```
edoardo: This is not passing `layerId` which results in `headersByLayer` to have `undefined` `layerId` and thus the `allHeaders` is always `null`.
This seems to cause the`Configure columns` button to always be disabled.

I didn't try with many visualizations but in all the ones I tried the button is disabled, so this seems to be a bug.

edoardo:
```suggestion
        onHeadersChange?.(headers, activeLayerId)
```

src/components/datatable/DataTable.jsx
```diff
@@ -425,12 +179,40 @@ const Table = ({
...
+    }, [onHeadersChange, headers])
```
edoardo:
```suggestion
    }, [onHeadersChange, headers, activeLayerId])
```

src/components/datatable/DataTable.jsx
edoardo: I tried these fixes locally and I got the configure columns popover working...

edoardo (commented):
I've looked through the code but it's a bit difficult to follow not knowing the app that much.
I've added a few comments here. This PR seem to have 1 bug in the Configure columns button, which I "fixed" locally to see how the configure menu works.

I've also poked around the Netlify a bit and have some questions I wrote in a Slack thread.

src/components/datatable/controls/ActiveLayerControl.jsx
BRaimbault: https://github.com/dhis2/maps-app/pull/3724/commits/dd2afd4756720795badd712df3d3a3fafc3f493b

src/components/datatable/DataTable.jsx
BRaimbault: Definitely a bug fixed downstream, I probably did not do enough testing of this PR alone.
Fixed here two with: https://github.com/dhis2/maps-app/pull/3724/commits/dd2afd4756720795badd712df3d3a3fafc3f493b

src/components/datatable/controls/ToolbarIconButton.jsx
BRaimbault: It's a bit of a preference call either way. Leaning toward keeping the wrapper since it ends up reused in a few more places as the stack progresses, but inlining would've been reasonable too.

BRaimbault:
Thanks for your review @edoardo, I addressed your specific comments and marked them as resolved. Also fixed the "Configure columns" bug you found (missing `activeLayerId` in `onHeadersChange`). Responded to the Slack thread items separately.

PR #3726, opened 2026-07-19
feat: extend data table support to TE, thematic, and event layers [PR5]

BRaimbault:

> **Diff size note:** this PR adds ~5,000 lines: ~1,900 are runtime code / ~3,100 (62%) are tests.

### Description

Extends the data table to the remaining layer types and rendering modes from the epic:

- **Tracked entity layers**: attribute columns + values now flow through the loader into the table (previously stripped)
- **Thematic layers with timeline or split-by-period rendering**: dynamic "current period" value column, plus any additional period addable as a side-by-side comparison column via the column picker
- **Server-side clustered event layers**: data table now opens on these too, with a "Show event details" in-table action for clustered rows
- **Facility and org unit layers**: Color/Icon/Group columns for layers styled by a group set
- **GeoJSON URL and tracked entity layers**: Color column
- **Event layers styled by a data item**: Legend/Range columns
- **Canonical column ordering** across all of the above: Legend → Range → Group → Color → Icon, previously inconsistent per layer type
- **Icon filter dropdown** now shows the actual icon thumbnail instead of the raw icon URL

Also included, surfaced while wiring up event layers:
- Server-side event clustering is now gated on the backend's advertised spatial support (`spatialSupport`), with a `forceClientCluster` escape hatch
- Fixes: mixed geometry/multi-geometry layers in the table, numeric `valueType` coercion for TE columns, a race when editing filters/columns while a server-clustered event layer's data is still extending, a resize-cancel bug, a crash + dead code path in `parseLayerConfig`, header alignment, and a data table performance pass

---

### Quality checklist

- [ ] Docs added _N/A_
- [ ] d2-ci dependencies replaced (https://github.com/dhis2/maps-gl/pull/659)

---

### Known issues

- [ ] Colorpicker popover position isn't app-controllable — all color swatches use a native `<input type="color">`; deferred to a follow-up
- [ ] Manual column-width resize (drag to widen/narrow a column) not included — deferred to a future PR
- [ ] Data table performance with very large (>100k row) event layers — highlighting/selection can take seconds; deferred, to be scoped separately


src/components/datatable/BottomPanel.jsx
```diff
@@ -51,25 +55,29 @@ const BottomPanel = () => {
...
-    const collapsedHeight = getCssVar('--data-table-controls-height')
-    const displayHeight = isCollapsed ? collapsedHeight : tableHeight
+    const { maxHeight, collapsedHeight, displayHeight } = getPanelHeights({
+        windowHeight: height,
+        dataTableHeight,
+        isCollapsed,
+        headerHeight: getCssVar('--header-height'),
+        toolbarHeight: getCssVar('--toolbar-height'),
+        controlsHeight: getCssVar('--data-table-controls-height'),
+    })
```
HendrikThePendric: Consider implementing a hook `usePanelHeights` that only takes `isCollapsed` (which is the only local state input for `getPanelHeights`.

src/components/datatable/DataTable.jsx
```diff
@@ -337,8 +373,184 @@ const Table = ({
...
+            visibleHeaders,
+            pinnedLeftOffsets,
+            pinnedColumnCount,
+            columnWidths,
+            columnOptions,
+            isAllSelected,
+            onToggleSelectAll,
+            headerRowRef,
+        ]
+    )
```
HendrikThePendric, line 532: I think extracting the JSX here into a `FixedHeaderRow` component and then returning that component would be a little bit cleaner. I see that this `useCallback` has 16 dependencies, so it won't be an ideal solution, because you'll end up with a component with many props and all of them still have to be a dependency of the `useCallback` too. But I suspect that quite a few of the current `useCallback` dependencies are unique to the header-row, so you should be able to make them internal to the new component, which means passing fewer props and fewer useCallback deps. 

In any case, I suspect this will help cleaning up this oversized component a bit.

src/components/datatable/DataTable.jsx
```diff
@@ -347,163 +559,11 @@ const Table = ({
...
-                                    </DataTableColumnHeader>
-                                )
-                            }
-                        )}
-                    </DataTableRow>
-                )}
+                computeItemKey={computeItemKey}
+                increaseViewportBy={VIEWPORT_OVERSCAN}
+                fixedHeaderContent={fixedHeaderContent}
                 itemContent={(_, row) => {
```
HendrikThePendric: Possibly this could also be extracted to a component that is returned by a useCallback hook

src/components/datatable/FilterInput.jsx
HendrikThePendric, line 1: I tried to add a comment to specific section of code here, but it's hard to tell which parts should or shouldn't be included in the selection, so I'll just settle for a general observation: This component has many computed variables and functions in its function body. Some are wrapped in useMemo and others are not. Quite a lot of the ones that are wrapped in `useMemo` are interdependent, I suspect some may even be purely intermediate variables. I think things could be cleaned up a bit by grouping things together. Not quite sure how the optimal solution would look, but it will be somewhere in between the current solution and a single call to useMemo that returns all local variables in its result object.

HendrikThePendric (dismissed):
LGTM, some minor code style comments.

src/components/datatable/BottomPanel.jsx
BRaimbault: Done, extracted into usePanelHeights(isCollapsed).
https://github.com/dhis2/maps-app/pull/3726/commits/20fba97625c82b8960bb550f49fcf06ded0d5393

src/components/datatable/DataTable.jsx
BRaimbault: Good point, but holding off on this one to avoid dealing with significant conflict.
Later PR7 (#3741) pulls the header row out into its own components (`SortableColumnHeader`/`SelectionCheckboxColumn`), for a different reason (reusing it in a new Combined table view), but effectively resolving your point.

src/components/datatable/FilterInput.jsx
BRaimbault: Also addressed by PR7 (#3741). It removes the exact block you're pointing at (the interdependent useSelector/filters/filterValue variables) and replaces it with props, to make the component reusable from the new Combined table.

PR #3648, opened 2026-04-26
feat: configurable decimal precision for automatic legends [PR4] [DHIS2-18242]

BRaimbault:
### Parent

- [DHIS2-18242](https://dhis2.atlassian.net/browse/DHIS2-18242)

### Implements

- [DHIS2-3156](https://dhis2.atlassian.net/browse/DHIS2-3156): Decimal places option for automatic legend classification
- [DHIS2-19984](https://dhis2.atlassian.net/browse/DHIS2-19984): Values not displayed for NUMBER data items with option sets in data table
- [DHIS2-19983](https://dhis2.atlassian.net/browse/DHIS2-19983): Range column not sorted numerically in data table
- [DHIS2-19982](https://dhis2.atlassian.net/browse/DHIS2-19982): Color picker button too narrow in option set and boolean styles
- [DHIS2-19203](https://dhis2.atlassian.net/browse/DHIS2-19203): Earth Engine style select uses "Steps" instead of "Classes"

### Overview

Fourth PR in the series extracted from the parent epic. Adds a `legendDecimalPlaces` setting that controls how many decimal places are shown in automatic legend range labels, and fixes four smaller issues surfaced during the same work.

#### Configurable decimal precision for automatic legends [DHIS2-3156](https://dhis2.atlassian.net/browse/DHIS2-3156)

A new `DecimalPlacesSelect` component (0–6 places) is added to the Classification and Single Color panels for automatic and single-color legends. The selected value is stored in Redux via a new `SET_LEGEND_DECIMAL_PLACES` action and persisted into `layer.config` JSON on save (via `favorites.js`) and parsed back on load (via `parseJsonConfig`).

`getLegendItems` accepts a `precision` option that rounds bin boundaries via `getRoundToPrecisionFn`. `getAutomaticLegendItems` accepts `legendDecimalPlaces` and stamps `decimalPlaces` onto each item. Both `LegendItemRange` and `Bubbles.jsx` read per-item `decimalPlaces` and pass it as `precision` to `formatWithSeparator`, so legend range labels and bubble text labels render with the configured precision. `styleByDataItem.js` propagates `legendDecimalPlaces` for event layer style-by-data-item.

`legendDecimalPlaces` is added to `validLayerProperties` so it survives the `pick` in `cleanMapConfig`.

Note that values are displayed using their stored precision; this setting only affects legend formatting. Also, range is only passed to the data table for Thematic Automatic and Prefefined legends, this behavior is not modified.

[...]

PR #3736, opened 2026-07-24
feat: add drill-down filters for data table date and org unit columns [PR6]

BRaimbault:

> **Diff size note:** this PR adds ~4,100 lines: ~1,800 are runtime code / ~2,300 (56%) are tests.

### Description

Replaces the flat value-list filter with a hierarchical drill-down popover for date/time and org unit data table columns:

- **Date/time columns**: filter by year → quarter/month → day → hour in an expandable tree, or type free text to match rows directly, same as before. "Any value"/"No value" pinned options carry over.
- **Org unit columns**: same drill-down experience, but through the org unit hierarchy (country → region → district → facility). Ancestor names for the tree are resolved via a new `useOrgUnitAncestorNames` hook.
- Both filters share a common `useGroupFilterInput` hook and `GroupFilterPopover` presentational component (extracted mid-PR from two near-duplicate implementations), plus a generic `prefixTree.js` utility for building/flattening/search-matching the tree.
- Org-unit-rendered columns that stay on the plain (non-tree) filter dropdown now resolve to their breadcrumb/own-name label instead of the raw id/path, and free-text search against them is name-aware.
- Sorting an org-unit-rendered column now sorts by the displayed name/breadcrumb rather than the raw path/id.

Also included, found while reworking org unit columns:
- Org unit id/name/level/path columns unified into a consistent shape across layer types (previously per-layer-type field names like `parentName`/`ouname`).
- Boolean-valued fields now render as Yes/No instead of raw `true`/`false`.
- Tracked entity layers: option-set-coded attribute values now resolve to their display name (previously showed the raw code); `createdAt`/`updatedAt` exposed as columns.

---

### Quality checklist

- [ ] Docs added _N/A_

---

### Screenshots

_supporting images_


src/hooks/useOrgUnitAncestorNames.js
```diff
@@ -0,0 +1,50 @@
...
+    const ids = useMemo(
+        () => [
+            ...new Set(
+                distinctPathValues.flatMap((path) =>
+                    String(path).split('/').filter(Boolean)
+                )
+            ),
+        ],
+        [distinctPathValues]
+    )
```
HendrikThePendric, line 28: I don't think this is correct — it will produce a set of org-unit IDs that includes plenty of items you already have the name for:
1. The last segment in the path is the unit's own ID, and in most cases you  already have its name.
2. Data-table units can be spread across levels (selecting `USER_ORGUNIT` together with its children/grandchildren, or a unit plus a level, are both normal selections), so an org-unit you already have can also be an ancestor of one of the other items. Those could be excluded too.
3. I also saw that the features already carry a `parentName` from geoFeatures? Not sure about this, but if so, the IDs of those named parents can be excluded as well.

I think to fix it you need to change the hook signature so it receives an array of org-units instead of just the paths. Then you iterate through them all and in a single pass you can create:
1. an incomplete `idToName` map, populated with all the units and parents you already have the name for
2. a set of `missingNameIds`

Then you fetch only those missing IDs, and once the results come in you can
populate the rest of the `idToName` map.

src/util/map.js
```diff
@@ -16,21 +20,19 @@ export const toGeoJson = (organisationUnits) =>
...
-            // Grand parent
-            if (isString(ou.pg) && ou.pg.length) {
-                const ids = compact(ou.pg.split('/'))
-
-                // Grand parent id
-                if (ids.length >= 2) {
-                    gpid = ids[ids.length - 2]
-                }
+            const ancestorIds =
+                isString(ou.pg) && ou.pg.length ? compact(ou.pg.split('/')) : []
```
HendrikThePendric, line 24: I think this is wrong. Check this URL: https://dev.im.dhis2.org/analytics-dev/api/organisationUnits?fields=id,name,path

What I am trying to say is that normally the `path` field's last segment is the unit's own ID. 

Maybe this different endpoint has a path that excludes it, but I'd find that surprising.

HendrikThePendric (changes requested):
I think there are some fundamental issues with the org-unit fetch side of this PR. Normally a `path` has the unit's own ID in it, but this PR seems to assume it stops at the parent.

src/util/map.js
BRaimbault: FYI the geoFeatures api deviates from standards with the pg/pi/pn (parent graph/id/name) props - see example here: https://dev.im.dhis2.org/analytics-dev/api/geoFeatures?ou=ou:LEVEL-4;O6uvpzGd5pu

The code in this PR (map.js's toGeoJson) appends ou.id to the pg-derived ancestors to build the full path, so I believe it's handling this correctly.

src/hooks/useOrgUnitAncestorNames.js
BRaimbault: Confirmed all three points and fixed it, reusing `buildKnownOrgUnitNames` from PR7 (#3648). `useOrgUnitAncestorNames` now takes a second argument, a pre-seeded `idToName` map built from each row's own `id`/`name` and `parentId`/`parentName`, and only fetches ids that aren't already known. Also added a module-level session cache, so a name fetched once is never fetched again for the rest of the session, even across unrelated table views.

https://github.com/dhis2/maps-app/pull/3736/commits/67dc16f93e50650f893d8fa3c563245ae2ce9262

BRaimbault:
Thanks for your review @HendrikThePendric, I addressed your specific comments and marked them as resolved. The `map.js` one turned out to be a `geoFeatures` API quirk rather than a bug, explained in the thread. The `useOrgUnitAncestorNames.js` one was a real inefficiency, fixed by reusing `buildKnownOrgUnitNames` from PR7 (#3648).

